PR #3, opened 2023-07-16
Archloom/Multiplatform Port

UnlikePaladin:
Supports and builds for Forge and Fabric
Pls review


skyrina:
read through the code and everything looks good, can someone verify that their avatar works correctly on both fabric and forge?
you can find download links to the jars [here](https://github.com/FiguraMC/Figura/actions/runs/5566040000)

Covkie:
java/org/moon/figura should be changed to java/org/figuramc/figura

UnlikePaladin:
I'll change it quickly before merging

PR #4, opened 2023-07-16
Enhance CI workflow, Update Gradle tools.

Kichura:
1. Changes the friendly names in workflow to look slightly better,
2. The chmod requirement will be merged into the build step as we do not need a seperate step for it specifically,
3. The java distribution will be replaced with temurin as microsoft's JDK basically is crap,
4. Replaces the irrelevant strategy matrix with a static JDK setup system as the workflow does not have any multiple JDKs in it,
5. Replaces the old quiltflower with the new vineflower dependency instead, This includes the two LuaJ libraries,
6. Updates websocket from 1.5.2 -> 1.5.3,
7. Updates Gradle from 8.1.1 -> 8.2.1.

Before merging, This PR should be tested before doing so to see if any problems occur during it - Suggestions are welcomed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,32 @@
-# NOTE: This readme file will be replaced soon.
+<h1 align="center"> Figura </h1>
+<p align="center"> We support Fabric Forge and Quilt! </p>
 
-# Figura
+```diff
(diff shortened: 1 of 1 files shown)


Covkie:

PR #7, opened 2023-07-16
Api additions again (previously KoTM/FRR#100 because github is stupid :> and KoTM/FRR#107 because :>)

lenrik1589:
1. Adds 3 events:
    - `PROJECTILE_RENDER`
    - `TRIDENT_RENDER`
    - `THROWN_ITEM_RENDER`
2. Adds 2 parent types
    - `TRIDENT`
    - `THROWN_ITEM`
3. Removes unnecessary "isLoaded" check every time `isModLoaded` is executed
4. Adds ability to un-set Action textures by
5. Adds extra data to `client.getDate()`
    - `timestamp`: time in millis since epoch
    - `time`: fully formatted time string
6. Adds ability to give pages mouse events and two new events:
    - `middleClick` fires for middle click mouse button
    -  `click` fires for every mouse button
7. Renames ActionWheelAPI.execute to ActionWheel.mouseClicked to be inline with mouseScrolled used for scrolling, this affects a likely small amount of scripts


Covkie:

lenrik1589:
@UnlikePaladin, could you please give pointers as to how it actually works now?

UnlikePaladin:
Sure, essentially you can't use or reference mod loader specific classes in the Common package, instead you point to a method of your creation that's annotated with @ExpectPlatform, then you create the impl (if you're using IDEA hovering over the method should provide a shortcut) for each mod loader.

UnlikePaladin:
For example in utils we have the PlatformUtils class which contains a isModLoaded method and other methods that reference mod loader classes directly, I also see you're doing some stuff with Mod metadata, for that there's the FiguraMetadata class which is meant to be a container/wrapper for the loader mod metadata, it's currently a bit basic but more methods can be added as needed. Creating and adding new Mixins or LuaJ stuff remains the same

fabric/src/main/java/org/figuramc/figura/utils/fabric/PlatformUtilsImpl.java
UnlikePaladin, line 1: I think this is redundant, we already have the FiguraMetadata Class, it just needs expanding with this info

lenrik1589: resolved by your commit, i guess

lenrik1589:
updated description to reflect the state of things this PR adds/changes

PR #8, opened 2023-07-16
Scoreboards

lenrik1589:
add host:getScoreboards and host:getObjectives
I think this is the correct stuff?


Covkie:

lenrik1589:
this one should be good to go

PR #9, opened 2023-07-16
Namespaced configs (semi-breaking change)

lenrik1589:
makes configs use their category in the their translation key and config json key
useful for any add-on that adds it's own config as categories are much less likely to conflict
avatars that use figura keys will error


Covkie:

lenrik1589:
same here

PR #10, opened 2023-07-16
add metatable handling to ReadOnlyLuaTables

lenrik1589:
makes AvatarAPI not ignore metatables
probably one change necessary would be making ReadOnlyLuaTable not strip userdata metatables as currently it breaks things like vectors, that are retreived from player's avatar variables


lenrik1589:
i forgot that i commited those changes, but it also removes redundant chunk is null checks, because (at least in 1.20.1) a chunk can not be null, and instead always returns a value from "empty" chunk

Covkie:

lenrik1589:
this one as well (though, original question about functionality still stands)

PR #21, opened 2023-07-18
Armor pivot support

omoflop:
Added support for armor pivots in BlockBench, so user's don't have to use code to manually offset armor anymore. Works with enchantment glints, trims, and dyes. 

Also added the pivots to the avatar wizard.

Finally, I also refactored a few random things

### DIFF
--- a/common/src/main/java/org/figuramc/figura/exporters/BlockBenchModel.java
+++ b/common/src/main/java/org/figuramc/figura/exporters/BlockBenchModel.java
@@ -6,6 +6,7 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 import org.figuramc.figura.math.vector.FiguraVec3;
 import org.figuramc.figura.math.vector.FiguraVec4;
+import org.figuramc.figura.model.ParentType;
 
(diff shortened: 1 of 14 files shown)


KitCat962:
have you tried using multiple pivots at the same time?
Does this function at all with modded armor (I assume it doesnt)

omoflop:
It should work fine with modded armor as long as it renders the same way vanilla does it (Which should be most) 
When using multiple pivot points it picks one to use to render. This is the same way the other pivot points work

KitCat962:
> When using multiple pivot points it picks one to use to render. This is the same way the other pivot points work

that is incorrect. All other pivot point parts render an additional copy of the thing. Having multiple RightItemPivot parented groups will render multiple rightItems

omoflop:
> > When using multiple pivot points it picks one to use to render. This is the same way the other pivot points work
> 
> that is incorrect. All other pivot point parts render an additional copy of the thing. Having multiple RightItemPivot parented groups will render multiple rightItems

Ok. I will look into this and make sure this works as intended. Also, closing this PR due to a few bugs that need fixed first.

PR #38, opened 2023-07-30
Add Optifine and Optifabric support

UnlikePaladin:
It was actually quite shrimple
This also fixes Oculus on Forge not being detected, as the mod id changes from iris to oculus.  

This is what was used to view Optifine's Source
[https://github.com/Cadiboo/OptiFineDeobf](https://github.com/Cadiboo/OptiFineDeobf)
It was created for use with pure Forge but because it remaps to Mojmap it works out. You just feed it an extracted Optifine Mod jar, (you can get this by opening the installer and pressing Extract instead of Install), you will also need a file that maps from SRG to Mojmap, then toggle 'Make Public' and 'Forge Dev jar and finally press Deobf
[NoCubes also has a guide for working alongside OF](https://github.com/Cadiboo/NoCubes#optifine-compatibility)

There is no actual dependency on Optifine itself so it is not required for compiling, it is optional and can be dropped in the libs folder, in the few places that OF's classes are referenced (just checking if shaders are on), it is done through Reflection so if the class is not present nothing happens.

Feel free to ask away if there's any questions, doubts or concerns.

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -40,8 +40,6 @@
     "immediatelyfast":  "*"
   },
   "breaks": {
-    "optifine": "*",
-    "optifabric": "*"
   },
   "custom": {
     "assets_version": "${assets_version}",
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -39,8 +39,6 @@
     "canvas": "*",
     "immediatelyfast":  "*"
   },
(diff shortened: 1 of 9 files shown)


fabric/src/main/resources/fabric.mod.json
```diff
@@ -40,8 +40,6 @@
     "immediatelyfast":  "*"
   },
   "breaks": {
```
Kichura: The breaks context can be removed entirely as there is nothing there.

Kichura (approved):

PR #39, opened 2023-07-30
Update Jitpack's JDK, Quilt Loader, Enhance comments.

Kichura:
1. Updates the quilt loader version to 0.20.0 beta to ship with certain patches along the way, *(Requires testing)*
2. The Jitpack JDK will be updated to 17.0.8+7 to address time-zone related bugs and that the JDK's code now defaults to C99 or later, *(+ also fixes the use of C23 attributes)*
3. A few comments have been enhanced to not look as messy - Also now clarifies where QSL and quilt loader updates are trackable. *(This part was done optionally although better than nothing)*

### DIFF
--- a/fabric/gradle.properties
+++ b/fabric/gradle.properties
@@ -1,4 +1,4 @@
 # loom.platform = quilt
 
 # Uncomment this ^ and change rootProject's "run_on_quilt" to true to run on quilt,
-# However it is not recommended building with this enabled however.
+# However building with this enabled is not recommended.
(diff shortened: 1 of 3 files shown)


fabric/gradle.properties
```diff
@@ -1,2 +1,4 @@
-# loom.platform=quilt
-# Uncomment this ^ and change rootProject's run_on_quilt to true to run on quilt, i wouldn't recommend building with this on
\ No newline at end of file
+# loom.platform = quilt
+
+# Uncomment this ^ and change rootProject's "run_on_quilt" to true to run on quilt,
+# However it is not recommended building with this enabled however.
```
UnlikePaladin: Grammar mistakes here, this should be 
```suggestion
# However building with this enabled is not recommended.
```

PR #44, opened 2023-07-31
Check for latest JDK, Validate Gradle distribution source, Final dependency patches.

Kichura:
1. Swaps the quotes with apostrophes instead although consider this a bit unnecessary of me,
2. The workflow will now check for a newer version of JDK 17 as github does not keep track of latest minor version, 
*(Default was 17.0.7 but we are on 17.0.8 now)*
3. The gradle wrapper jar will now also be checked to see if it's distribution source is properly signed,
4. This PR i will make will contain the last dependency patches i will be personally doing - LuaJ and websocket dependencies in properties file however are now also documented.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,5 @@
 name: Build Figura using Gradle
 on: [pull_request, push]
-
 jobs:
   build:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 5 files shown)


Kichura:
Updated the title of this PR to include one last patch-list of dependencies cuz better than never doing it.

PR #48, opened 2023-08-02
Animated emojis, armor crash fix, json apis

omoflop:
title


Covkie:
If anyone has no objections ill merge this

UnlikePaladin:
Why is this removing the GoL easter egg without mentioning it anywhere?

UnlikePaladin:
If anything debug mode should be a regular config option on the settings screen and not be hidden behind a key combo that is famously used for easter eggs

fabric/src/main/java/org/figuramc/figura/config/fabric/ModMenuConfig.java
```diff
@@ -3,11 +3,12 @@
...
 import com.terraformersmc.modmenu.api.ModMenuApi;
 import org.figuramc.figura.FiguraMod;
+import org.figuramc.figura.config.Configs;
 import org.figuramc.figura.gui.screens.ConfigScreen;
 
 public class ModMenuConfig implements ModMenuApi {
     @Override
     public ConfigScreenFactory<?> getModConfigScreenFactory() {
-        return parentScreen -> new ConfigScreen(parentScreen, FiguraMod.DEBUG_MODE);
+        return parentScreen -> new ConfigScreen(parentScreen, Configs.DEBUG_MODE.value);
```
UnlikePaladin: You forgot to change this to your FiguraMod.debugModeEnabled()

common/src/main/java/org/figuramc/figura/gui/screens/AbstractPanelScreen.java
```diff
@@ -216,7 +216,8 @@ public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
         egg += (char) keyCode;
         egg = egg.substring(1);
         if (EGG.equals(egg)) {
-            Minecraft.getInstance().setScreen(new GameScreen(this));
+            Configs.DEBUG_MODE.setValue("true");
```
UnlikePaladin: Might as well make it a regular option on the screen, it doesn't need to be hidden behind a key combo. We already expose several options normal users don't mess with like the ip backend option.

common/src/main/java/org/figuramc/figura/FiguraMod.java
```diff
@@ -49,6 +48,13 @@ public class FiguraMod {
     public static boolean parseMessages = true;
     public static boolean processingKeybind;
 
+    public static boolean debugModeEnabled() {
+        if (Configs.DEBUG_MODE == null) return false;
+        Boolean b = Configs.DEBUG_MODE.value;
+        if (b == null) return false;
+        return b;
```
UnlikePaladin: you do know you could simplify this entire method to
```suggestion
   return b == true;
```
lol

UnlikePaladin (dismissed):
The rest of the changes relating to animated emojis look good to me, just these minor things.

omoflop:
@UnlikePaladin thanks for the feedback, will fix all that stuff 👍

PR #52, opened 2023-08-08
Data related features

lexize:
This pull requests adds some data related features like FileAPI. Full list of changes:
### File API
API made to provide features to work with files.
Operations related to files can be done only in `figura/data` folder an only on host.
### Data API
This API adds features related to advanced data manipulation, as input/output streams, data readers for input streams, data providers for output streams, and also byte buffers.
### Json API
API that adds some types and functions for JSON serialization and deserialization.
### Resources API
Allows you to store raw file data from your avatar folder in avatar itself. To include resource in your avatar you must add resource path/[Glob  pattern](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob) to `resources` array in your metadata
### Networking API
Everything related to networking. Right now it is only HTTP requests, TCP is planned.
To use Networking API you will need to enable it in settings and in avatar permissions, and also setup your Network Filter, by setting restriction mode to "Whitelist", "Blacklist" or "None" and filling your filter with links and matching rules.
Networking is being logged to `figura/logs`, you can toggle logging off in settings.

#### Config:
Added an interface named SerializableConfig, that provides methods for direct serialization and deserialization from JsonElements
#### Lua Runtime
Changed `pairs` and `ipairs` functions to make them able to be work with userdata, and not only tables
#### ConfigAPI
ConfigAPI now uses `figura/config` folder instead of `figura/data`. `figura/data` is now used by FileAPI


lexize:
:blobfoxfloofcofe:

Covkie:
> should be merged when backend v3 will be too

-skylar

lexize:
Im sorry for force pushes but my git (or maybe this is github problem) refuses to push commit because blah blah your branch is older than remote branch, while it is not. Im so tired of this so im just force pushing shit

Covkie:
I don't know if this is an issue with me not knowing how to navigate the mobile UI but it seems the en_us.json entries for this pr were lost with the force push?

lexize:
No i just removed the en_us changes from pr because if i would try to make them match current en_us formatting I would die

common/src/main/resources/assets/figura/lang/en_us.json
```diff
@@ -617,6 +620,7 @@
     "figura.docs.globals.pings": "The global instance of PingAPI",
     "figura.docs.globals.textures": "The global instance of the TextureAPI",
     "figura.docs.globals.config": "The global instance of the ConfigAPI",
+    "figura.docs.globals.net": "The global instance of the NetworkingAPI",
```
Covkie: This does not have any use and I assume you forgot to remove it in commit `30ef162f`(No networking 😔). I included it for posterity's sake.

Covkie (dismissed):

common/src/main/resources/assets/figura/lang/en_us.json
lexize: Wait why tf is it here? :skull: 
As i remember i literally made my pr use en_us from main branch

Edit: Yeah this en_us is from old commit, it shouldnt be here at all.

Covkie: I took the latest commit still with the en_us translations that you made and modified the entries to be compatible with the current format so the doc entries and everything else are translatable.

lexize: Ah, ok, thank you. Yes this thing is not used anymore as i removed everything related to Networking, let me remove it rq

Issue #50, opened 2023-08-02
semi-frequent crash


kcin2001:
with seemingly no cause i semi frequently crash in both multiplayer and more rarely singleplayer
attached is the latest instance of it but i have like 20 of these, always the same error
[crash-2023-08-02_12.54.23-client.txt](https://github.com/FiguraMC/Figura/files/12244160/crash-2023-08-02_12.54.23-client.txt)

KabanFriends:
This happens to me as well
For me it frequently happens in Multiplayer

purpledeni:
Same for me

KitCat962:
from what I can tell from the crash report, someone forgot to do an avatar null check when rendering players offscreen

UnlikePaladin:
Will look into it

KabanFriends:
Downgrading to Sodium 0.4.10 fixes the crash
So this is a Figura's incompatibility with Sodium 0.5.0

kcin2001:
its definately not just sodium as i lack that mod

PR #63, opened 2023-08-17
Fix issue #50

alekso56:
untested, but i believe this was intended to be a nullcheck for both avatar and profile, not a set for avatar.


UnlikePaladin:
Not sure about this, I believe this IS an intentional set for avatar, it uses a ternary to check that the game profile isn't null and if so it gets the avatar using AvatarManager.getAvatarForPlayer(profile.getId()), changing this to check that avatar isn't null would mean avatar never even gets set in the first place, which would result in the call to AvatarManager literally being purposeless which would introduce more issues. This needs thorough testing.

alekso56:
> Not sure about this, I believe this IS an intentional set for avatar, it uses a ternary to check that the game profile isn't null and if so it gets the avatar using AvatarManager.getAvatarForPlayer(profile.getId()), changing this to check that avatar isn't null would mean avatar never even gets set in the first place, which would result in the call to AvatarManager literally being purposeless which would introduce more issues. This needs thorough testing.

after re-reading like 3-4 times i believe you are correct, i don't think a ternary should be used here, because i can't read it.
ill investigate the issue further before making further comments. thank you.

PR #65, opened 2023-08-20
Attempt 2 (to) fix  issue #50

### LINKED ISSUE #50: semi-frequent crash

alekso56:
This time i got 3  players who experienced the issue constantly because i myself couldn't recreate the issue.

The issue is a race condition with avatar being set to null in a render  at line 56 while at the  same time being called for permissions in line 92.
(The slower the computer  the  better for  recreation) 

i do not  know why this happens in the chunkbuilder. but i believe it is related to a bigger issue in the chunk thread builder renderer.
removing the nullset fixed all 3 players.  
This time the code is tested and found  to be working. closes #50

### DIFF
--- a/common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
@@ -53,7 +53,7 @@ public abstract class SkullBlockRendererMixin implements BlockEntityRenderer<Sku
 
         // avatar
         Avatar localAvatar = avatar;
-        avatar = null;
+        SkullBlockRendererMixin.avatar = null;
 
         if (localAvatar == null || localAvatar.permissions.get(Permissions.CUSTOM_SKULL) == 0)
             return;
@@ -89,11 +89,24 @@ public abstract class SkullBlockRendererMixin implements BlockEntityRenderer<Sku
 
     @Override
     public boolean shouldRenderOffScreen(SkullBlockEntity blockEntity) {
-        return avatar == null ? BlockEntityRenderer.super.shouldRenderOffScreen(blockEntity) : avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1;
+    	// if we have an avatar, check permissions
+    	if(SkullBlockRendererMixin.avatar != null  && SkullBlockRendererMixin.avatar.permissions != null) {
+    		if(SkullBlockRendererMixin.avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1) {
+    		    return true;
+    		}else {
+    			return false;
+    		}
+    	}
+    	//no avatar present, default rendering used.  
+    	return BlockEntityRenderer.super.shouldRenderOffScreen(blockEntity);
     }
 
     @Inject(at = @At("HEAD"), method = "getRenderType")
     private static void getRenderType(SkullBlock.Type type, GameProfile profile, CallbackInfoReturnable<RenderType> cir) {
-        avatar = profile != null && profile.getId() != null ? AvatarManager.getAvatarForPlayer(profile.getId()) : null;
+    	// set skull owner's avatar to be associated with this skull.
+    	SkullBlockRendererMixin.avatar = null;
+    	if(profile != null && profile.getId() != null) {
+    		SkullBlockRendererMixin.avatar = AvatarManager.getAvatarForPlayer(profile.getId());
+    	}
     }
 }
--- a/common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
@@ -88,15 +88,6 @@ public abstract class SkullBlockRendererMixin implements BlockEntityRenderer<Sku
 
     @Override
     public boolean shouldRenderOffScreen(SkullBlockEntity blockEntity) {
-    	// if we have an avatar, check permissions
-    	if(avatar != null  && avatar.permissions != null) {
-    		// if we have permissions to render, then we render, else we don't render.
-    		if(avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1) {
-    		    return true;
-    		}
-    		return false;
-    	}
-    	//no avatar present, default rendering used.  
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
```diff
@@ -89,11 +88,25 @@ public void render(SkullBlockEntity skullBlockEntity, float f, PoseStack poseSta
...
+    	}
+    	//no avatar present, default rendering used.  
+    	return BlockEntityRenderer.super.shouldRenderOffScreen(blockEntity);
     }
 
     @Inject(at = @At("HEAD"), method = "getRenderType")
     private static void getRenderType(SkullBlock.Type type, GameProfile profile, CallbackInfoReturnable<RenderType> cir) {
-        avatar = profile != null && profile.getId() != null ? AvatarManager.getAvatarForPlayer(profile.getId()) : null;
+    	// reset avatar for skull
+    	avatar = null;
```
UnlikePaladin: This whole thing can be simplified to
```suggestion
    	avatar = profile != null && profile.getId() != null ? AvatarManager.getAvatarForPlayer(profile.getId()) : null;
```
There is literally no reason to expand the logic, the ternary is simpler and more compact, if you don't understand these you might want to review on some of the Java basics again because they are a useful tool and a basic operation. This change is unnecessary, the behavior is the exact same.

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
```diff
@@ -89,11 +88,25 @@ public void render(SkullBlockEntity skullBlockEntity, float f, PoseStack poseSta
...
+    	// if we have an avatar, check permissions
+    	if(avatar != null  && avatar.permissions != null) {
+    		// if we have permissions to render, then we render, else we don't render.
+    		if(avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1) {
+    		    return true;
+    		}
+    		return false;
+    	}
+    	//no avatar present, default rendering used.  
+    	return BlockEntityRenderer.super.shouldRenderOffScreen(blockEntity);
```
UnlikePaladin: Again this whole thing can be simplified to
```suggestion
    	return avatar == null || avatar.permissions == null ? BlockEntityRenderer.super.shouldRenderOffScreen(blockEntity) : avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1;
```
Use ternaries to your advantage. Also ```avatar.permissions.get(Permissions.OFFSCREEN_RENDERING) == 1``` is already a boolean, there is no reason to expand the logic there either. Something tells me you really should review the basics, no offense meant, but we like to keep things simple and compact. Also the extra comments aren't necessary or wanted when the code is already self explanatory.

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
```diff
@@ -51,9 +51,8 @@ private static void renderSkull(Direction direction, float yaw, float animationP
         SkullBlockRendererAccessor.SkullRenderMode localMode = SkullBlockRendererAccessor.getRenderMode();
         SkullBlockRendererAccessor.setRenderMode(SkullBlockRendererAccessor.SkullRenderMode.OTHER);
 
-        // avatar
+        // avatar pointer incase avatar variable is set during render. (unlikely)
         Avatar localAvatar = avatar;
-        avatar = null;
```
UnlikePaladin, line 56: You might want to read the comments from the bottom up, I started reviewing there so it might make more sense.

Not really sure about removing this here. This setting to null is intentional, and removing it is not a fix for anything. The crash was happening because the permissions field was null, which is now accounted for in the new check in shouldRenderOffScreen which should fix the issue itself. Did you try testing without changing this here and only updating the problematic ternary/if and seeing if that fixed the problem?

UnlikePaladin (changes requested):
Might want to read the comments from the bottom up, the PR is appreciated tho

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
alekso56: During testing avatar was  null, the permission field was null because the avatar field was null.
"java.lang.NullPointerException: Cannot read field "permissions" because "net.minecraft.class_836.avatar" is null"
We still encountered crashes during testing with just the permissions check.

If you want to keep the set here, i would insist on using a lock or similar during operations on it.
ex `private final AtomicReference<Avatar> avatar = new AtomicReference<Avatar>();`

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
alekso56: I was always taught that if you are doing more than one if condition in a variable setting ternary you should expand the logic or use braces  to make the code easier for other developers to  read. I'll change it to a ternary with braces if  that's ok?

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
alekso56: I guess code tends to expand during debugging, I'll remember that.

common/src/main/java/org/figuramc/figura/mixin/render/renderers/SkullBlockRendererMixin.java
UnlikePaladin: I see, what about making a copy in shouldRenderOffscreen instead? Did you try that?

alekso56: No, i did not try that. It seems odd to use that as a fix, but i guess it would technically work. 
I'll assemble the team and give it a shot.

alekso56: It seems to work. pushed changes here.

UnlikePaladin: Great, merging now, thanks for the PR :)

PR #66, opened 2023-08-20
Update README.md

ChloeSpacedOut:
this was overlooked >~>


Kichura (approved):
good catch. (btw commit should've referenced the typo fix instead of being generic)

PR #67, opened 2023-08-21
ModelPart Api extension

omoflop:
+ Added **getPrimaryTexture** and **getSecondaryTexture** 
    + Both return the respective texture override type, and the value (if any)
+ Added **getPrimaryDefinedTexture** and **getSecondaryDefinedTexture**
    + Takes an integer as a parameter, and returns a table containing the texture data of the specified face


omoflop:
i clearly dont know what im doing because i did not mean to close this

PR #69, opened 2023-08-23
Fixes ImmediatelyFast breaking animated emoji rendering

UnlikePaladin:
Old resources are now deleted automatically when they aren't present upstream on the backend.
Added a method to get a specific mod's version to PlatformUtils


ElectricDaFearox:
just wanted to say

Nice

PR #73, opened 2023-08-24
remove hearts and add emoji category names

aprilkenopsia:
free me cookie

pls feedback on emoji category names as i am unsure if they are good enough


UnlikePaladin:
Looks good to me

PR #75, opened 2023-08-28
Support ImmediatelyFast's new API

UnlikePaladin:


gradle.properties
```diff
@@ -31,7 +31,7 @@ fabric_loader_version = 0.14.22
 # https://modrinth.com/mod/iris
 modmenu = 7.2.1
 iris = 1.6.4+1.20
-
+immediately_fast= 1.2.0+1.20.1
```
Kichura: The equal sign isn't consistent with rest of properties + Please include source of mod.

Kichura (commented):
lgtm, although one tiny inconsistency is present.

UnlikePaladin:
that literally is so minimal, plus it has no effect but whatever

PR #90, opened 2023-10-01
Made text clickable in the motd

omoflop:
Text can now have click and hover events in the MOTD.
Only SHOW_TEXT and OPEN_URL events work, as I had to manually implement them, and the others aren't necessary.

Also fixes the bug where the avatar info panel would overlap with the MOTD.


UnlikePaladin:
Looks good to me, merging now

PR #92, opened 2023-10-01
Add `world.isChunkLoaded()` to check if chunks are loaded

PoolloverNathan:
Also makes it so that `world:getBlockState()` properly returns VOID_AIR instead of AIR, and documents that so nobody can change it

Cherrypicked-from: 421dbc43d88ad62725e439b8f94f769991f30625

---
Replace-PR: #88
Request-Review: @UnlikePaladin


UnlikePaladin:
Lgtm

PR #96, opened 2023-10-03
Added re-colorable emojis

omoflop:
Emojis can now include the field "color", which takes a hex string. When this field is present, it allows the emoji to be recolored, and when no color is given, it uses it's default, given by the "color" field.


Covkie:
YIPPPEEE ANOTHER FLOPPER W

ElectricDaFearox:
its all fun and games until it breaks with 50+ mods we don't know about.

UnlikePaladin:
this cannot possibly go wrong

PR #99, opened 2023-10-12
new function: ClientAPI.generateUUID()

KitCat962:
Adds a new function which generates a random uuid.

Odd how we didnt have this before.


UnlikePaladin:
No reason we can't sneak this one in last minute :p

PR #101, opened 2023-10-19
Fix PlatformUtils crashing the game whenever a version of a mod from a newer version of MC is present

superpowers04:
Instead of checking for the current minecraft version, this'll split the version by [+_,] and only use the first result. As a side effect, this'll fix mods that use a +, _ or , in their version for whatever reason


superpowers04:
Ignore this being a merge from the Extura repo, since I can only have one fork of a repo, it used that repo

PR #106, opened 2023-10-28
create sync workflow

repository:
syncs strings from `en_us.json` to assets repo

### DIFF
--- a/.github/workflows/update-base.yaml
+++ b/.github/workflows/update-base.yaml
@@ -1,0 +1,38 @@
+name: Update Base Strings
+
+env:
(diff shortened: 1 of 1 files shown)


Covkie:
shouldn't the branch be 'HEAD' instead of '1.20' to prevent future changes needing to be made when the head branch changes

KitCat962:
1.20 is the head/default/main branch

repository:
> shouldn't the branch be 'HEAD' instead of '1.20' to prevent future changes needing to be made when the head branch changes

that doesn't work for specifying branches in the push event

Covkie:
> that doesn't work for specifying branches in the push event

ah okay I didn't know that.

PR #109, opened 2023-10-30
Trident Parent Type And Event

Manuel-3:
Adds a new ParenType `TRIDENT` and a `TRIDENT_RENDER` event, both of which work exactly like the `ARROW` and `ARROW_RENDER` events, except that the model is not being rotated into the direction it is flying. (Which I think `ARROW` ParentType does do, right?)

Here is an example Avatar that uses these features to replace the vanilla Trident model and also rotates the model to face into the flying direction: [Direct Link](https://cdn.discordapp.com/attachments/1152237863468007454/1152237900608585749/trident-example-avatar.zip) or [Discord Message](https://discord.com/channels/1129805506354085959/1152237863468007454/1152237901053165568)

### DIFF
--- a/common/src/main/resources/figura-common.mixins.json
+++ b/common/src/main/resources/figura-common.mixins.json
@@ -4,6 +4,7 @@
   "package": "org.figuramc.figura.mixin",
   "compatibilityLevel": "JAVA_17",
   "mixins": [
+    "render.renderers.TridentRendererMixin"
   ],
   "client": [
     "BiomeAccessor",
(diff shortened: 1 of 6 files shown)


common/src/main/resources/figura-common.mixins.json
```diff
@@ -4,6 +4,7 @@
   "package": "org.figuramc.figura.mixin",
   "compatibilityLevel": "JAVA_17",
   "mixins": [
+    "render.renderers.TridentRendererMixin"
```
UnlikePaladin: Please remove this, we don't register our mixins on the common side, only on the client, you have it here and down with the client stuff as well, with this removed the rest looks good to me, so it could still be merged in time for 0.1.3

PR #112, opened 2023-10-31
Add comment regarding jitpack's installation

Kichura:
Indicates what the jitpack script does, not much else to add here in terms of description.


Kichura:
To be redoed. *(as certain elements were missing, also cuz of branch name)*

PR #124, opened 2023-11-25
Adds additional functionality to raycasting functions

KitCat962:
Adds additional functionality to the raycastBlock and raycastEntity functions.

For raycastBlock, the function signature has been changed to `raycastBlock(String blockCastType, String fluidCastType, Object x, Object y, Double z, Object w, Double t, Double h)`

raycastBlock now takes in the BlockRaycastType, and FluidRaycastType, which more accurately defines how the raycast behaves.

For example, specifying `"COLLIDER"` will use the collision shape for the raycast.

When `blockCastType` is nil, it will default to `"OUTLINE"`. When `fluidCastType` is nil, it will default to `"NONE"`.

A `block_raycast_types` and `fluid_raycast_types` have been added to the enum docs, and the documentation of `raycastBlocks` has been updated.

For raycastEntity, the function signature has been changed to `raycastEntity(LuaFunction predicate, Object x, Object y, Double z, Object w, Double t, Double h)`

raycastEntity now takes in a predicate function which will filter out unwanted entities.

For example, the following will filter out the owner of the avatar when raycasting, which cannot be done without a predicate
```lua
local playerEyeHeight = player:getPos():add(0, player:getEyeHeight(),0)
print(world:raycastEntity(
  function(entity) return entity:getUUID() ~= player:getUUID() end, 
  playerEyeHeight, 
  playerEyeHeight + player:getLookDir() * 4
))
```

Unfortunately, because of the need of the avatar owner for userdata conversion, I needed to make the WorldAPI an instance instead of static, meaning that all `WorldAPI` functions now require `:` to execute correctly.

This is a horrible solution, so I wish to discuss with the devs about alternative solutions.

I currently have 3 solutions: 
  1. Move the raycast functions to an API that already is an instance, or create a new API.
  2. Make all WorldAPI functions static, *except* for the raycast functions.
  3. Because Figura is still in beta, don't worry about the massive API change.


KitCat962:
currently the dif is fucked because of crlf. Let me fix that

PR #129, opened 2023-12-03
Fix incorrect uv calculations due to BlockBench 4.9

KitCat962:
Blockbench 4.9 adds per-texture uv sizes which completely breaks how Figura calculates uvs.
This fixes that by grabbing the texture width, height, uv_width, and uv_height from the bbmodel and using those to calculate the uvs per texture, instead of the ProjectUV

If the texture width doesnt exist, it is assumed that the bbmodel is from 4.8 or below and will use the old method for backwards compatibility


UnlikePaladin:
Thanks for the fix!

PR #131, opened 2023-12-06
*raycasting*

KitCat962:
Moves raycasting functions to it's own RaycastAPI so that we can do predicate.
`raycastBlock` has been renamed to `block` and now takes in a BlockRaycastType and FluidRaycastType
`raycastEntity` has been renamed to `entity` and now takes in a predicate function with a single Entity argument and returns a boolean. Controls which entities are valid returns when raycasting

adds the BlockRaycastType and FluidRaycastType enums to the docs.


UnlikePaladin:
Looks good to me!

PR #133, opened 2023-12-11
Optimize Avatar NBT - Scripts

KitCat962:
Optimizes the Avatar nbt by saving scripts as a Compound Tag mimiking the file structure instead of repeating the path per script.
Additionally, it updates the `/figura debug` command to reflect these changes

This is the nbt pre-optimization
image

This is the nbt post-optimization
image

While this doesn't save much filesize, it doesnt cost any filesize meaning this is a free optimization.

This will break avatars that rely on the old nbt format via `avatar:getNBT()`, but in my opinion, ".moon files are an internal format liable to change" :P

`require` functionality is untouched

**edit** I switched the images lmao


KitCat962:
#135 Implements these changes. Close when #135 is merged

PR #147, opened 2023-12-24
Fixed vanilla inventory behaving odd due to previous pr

KitCat962:
*sigh*
look, I do test my prs. Seriously I do.
and for some reason there is *always* this one single trivial thing that somehow goes under the radar.

Anyways, this pr fixes the vanilla inventory being jank from pr #144 and cleans up `UIHelper.drawEntity` a bit.
For some reason, pitch and yaw were being interpreted as mouse coordinates depending on the context. While it made the mixin clean, it made the drawEntity function arguments worse. The pitch and yaw arguments are now always pitch and yaw, regardless of the context.


UnlikePaladin:
lgtm

PR #151, opened 2023-12-26
Raycast vector arguments are now at the beginning of the function

KitCat962:
Changed the Raycast functions so that the vectors are at the beginning of the function signature. This is what I wanted to do originally, but didnt have the tech to take into account arguments after the vectors.

Additionally, this adds a generic vector parsing util function that can handle everything that all the current vector parsing functions can do, but in a single function. The pre-existing hardcoded functions will definitely be faster than this new function, but this is the only function that allows for parsing vectors with arguments after the vectors, so use it whenever another raycast function situation happens again.


KitCat962:
oops. I committed to the wrong branch. Oh well. There is now a new raycast function that raycasts user provided AABBs using vanilla functions

UnlikePaladin:
because the raycast api hasn't actually made it's way into a stable release, this is okay, it looks good to me.

PR #152, opened 2023-12-28
Redirects prereleases to github releases

KitCat962:
As we are now releasing prereleases to the github, the "update version" redirect now points to github instead of modrinth, but only when the latest version is identified as a prerelease. Normal relreases are still redirected to modrinth.

Additionally, this fixes a bug where the regex expects the build metadata to begin with a `+`, when the build metadata does not have a `+`. Build metadata was leaking into prerelease data, so this fixes a bug related to version parsing.

Also, I made the version regex span multiple lines so that it will be easier to read in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ allprojects {
 
 	java_version = rootProject.java_version
 	archivesBaseName = rootProject.archives_base_name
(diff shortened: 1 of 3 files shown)


UnlikePaladin:
I hate regex lol, thanks!

PR #153, opened 2023-12-29
Split world.getTimeOfDay into world.getTimeOfDay and world.getDay

TheBunnyMan123:
Currently, `world.getTimeOfDay` outputs the total time (the value in the F3 menu, not the game time), which is misleading compared to what the function is called (doesn't actually get the time of the current day). 

This splits world.getTimeOfDay into the following functions:
```lua
world.getTimeOfDay()
    --Gets time of current day (0 ... 24000)
world.getDay()
    --Gets current day (0 ...)
```


UnlikePaladin:
Yeah this one has always been weird to me too, but we like to avoid breaking existing avatars when possible, it's kind of the point of having an API, so it'd be great if you could change getTimeOfDay to be a new method that doesn't replace the existing one, or provide a new method with the old behavior with better naming, either would be fine.

UnlikePaladin (changes requested):
refer to the last comment for more info, but we do prefer to not break or change existing methods from under the user's feet without telling them once things have existed for a while as we do want to provide a stable API

TheBunnyMan123:
Alright, I'll get working on this when I have access to my computer

TheBunnyMan123:
I have done so and tested! (it works on my machine)

Basically I re-instated the old getTimeOfDay and renamed the new one to getDayTime

UnlikePaladin:
lgtm

UnlikePaladin (approved):

PR #155, opened 2023-12-31
HostAPI Avatar Loading

Manuel-3:
Adds functionality to HostAPI to load avatars locally as well as upload them. This for example allows Origins avatars to be split apart and be loaded by one main avatar.

```json
"figura.docs.host.loadLocalAvatar": "Loads an avatar locally. If path is empty or nil, it will instead unselect the avatar you are currently wearing.",
"figura.docs.host.loadAndUploadAvatar": "Loads an avatar and then uploads it to the cloud. If path is empty or nil, it will instead clear your avatar from the cloud and unselect the avatar you are currently wearing.",
"figura.docs.host.uploadThisAvatar": "Uploads the currently equipped avatar to the cloud.",
"figura.docs.host.clearCloud": "Clears your uploaded avatar.",
"figura.docs.host.reloadFromCloud": "Downloads and equips your avatar that is currently on the cloud.",
```


skyrina:
this feature is out of scope. uploading avatars is an expensive task which will be heavily ratelimited in the future and will soon be obsolete due to multi avatar support coming in the future

Manuel-3:
How exactly is "avatar uploading" obsolete just because you will be able to upload multiple in the future? It can just be adjusted to upload one avatar at a time.

Also, rate limiting is perfectly fine and doesn't hinder this feature at all. In fact its a good thing, because it prevents people from accidentally or intentionally spamming it.

Not sure what you mean by it being out of scope, since it is clearly possible with just a few lines of code.

PR #158, opened 2024-01-04
Still allows an avatar to load if none is found for player

jasperalani:
Allows function to continue instead of returning the result of getAvatarForPlayer


UnlikePaladin:
please target the 1.20 branch, but yeah this seems good to me

UnlikePaladin (changes requested):
just change the target branch to the correct one, other than that, lgtm

jasperalani:
oop i think i messed it up
was that the right thing to do?

PR #159, opened 2024-01-07
Client.getEnum

4P5:
This PR adds `client.getEnum`, which takes a string and returns a list of values for the Figura enum with that name.

For example, `client.getEnum("render_modes")` returns the table shown in the attached image.

Feel free to close if this change is too intrusive (I think it might break encapsulation for FiguraListDocs?), I don't see a huge use for this feature but it would be nice to have!

java_YNcC09xuA4


UnlikePaladin:
looks fine to me tbh, don't know what the uses would be but it doesn't look that intrusive to me

PR #160, opened 2024-01-08
client.getRegistry() and world.getHeight()

4P5:
This PR adds two new methods, `world.getHeight` and `client.getRegistry`, and two new enums, `HEIGHTMAP_TYPES` and `REGISTRIES`.

`world.getHeightmap()` takes a Vec2 or x/z coordinates, and a string representing the type of heightmap. It returns the world height at the given position, based on the heightmap string.

`client.getRegistry()` takes a string from the `REGISTRIES` enum, which is all the vanilla registries (stuff like blocks, items, particle types, cat variants, villager types). It returns a numbered list of the strings in the registry. For example, `client.getRegistry("block")` would return a list of all block IDs defined in the client's registries.


UnlikePaladin:
looks good to me!

PR #167, opened 2024-01-15
Use JDK 17 instead of 21 in terms of codespaces

Kichura:
Specifies JDK 17.0.10 *(temurin)* to be used instead of jitpack just fetching the actual latest it can find which is 21 since that doesn't really work well in certain repos such as this one.


UnlikePaladin:
Forge has some problems with Java 21, so thanks for noticing/fixing this!

PR #170, opened 2024-01-23
Cancel sounds

KitCat962:
This pr allows preventing any sound from playing when a user returns `true` in the `ON_PLAY_SOUND` event.
Any player can block any sound, so this is only enabled by default for HIGH trust and above.
users can use the sound's position relative to their player's position to target sounds their player generates.

I had 4p5 bugtest this time, so there should be no issues

This completes the [Allow canceling any vanilla sounds using the ON_PLAY_SOUND event](https://discord.com/channels/1129805506354085959/1174799238710960168) suggestion from discord


KitCat962:
do not feel obligated to include this for 0.1.3. It can wait

UnlikePaladin:
looks good to me

PR #172, opened 2024-01-25
EntityTasks

applejuiceyy:
entity tasks or something
image


KitCat962:
can you give the entity rendertask a context for the render event?

applejuiceyy:
I have the slight idea that they shouldn't render through figura, I remember enabling it for player skulls in item tasks and ending up opening a giant can of worms

KitCat962:
can I suggest adding a function to allow generating an Entity copy from an existing entity?

Also, how do you use the setNbt function? What is the format of the nbt it is expecting?

applejuiceyy:
just the usual nbt string, like the summon command

KitCat962:
are you sure about that?
errors with or without the `{}`

image

also yea vm error bad

PoolloverNathan:
> are you sure about that?
> errors with or without the `{}`
> 
> image
> 
> also yea vm error bad

@KitCat962 IIRC from reading the source that one time, it's the entity's serialized NBT (JSON containing `id` property and other entity properties next to it)

KitCat962:
Also, none of the functions added to EntityTask return self. They should return self to match all other RnederTask functions

applejuiceyy:
> are you sure about that?
> errors with or without the `{}`
> 
> image
> 
> also yea vm error bad

no just the nbt part

applejuiceyy:
> Also, none of the functions added to EntityTask return self. They should return self to match all other RnederTask functions

yes, calm down, I only have 4 hours per day and not all of them are in intellij

KitCat962:
> > Also, none of the functions added to EntityTask return self. They should return self to match all other RnederTask functions
> 
> yes, calm down, I only have 4 hours per day and not all of them are in intellij

sorry. I got too excited. I can't tell what is a "i forgor" and what is a "I know about this and I'll do it later". Usually prs are only put in draft when changes are still being done, so since this was an open pr I assumed what I saw was going to be "final".

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1303,6 +1302,19 @@ public SpriteTask newSprite(@LuaNotNil String name) {
         return task;
     }
 
+    @LuaWhitelist
+    @LuaMethodDoc(
+            overloads = @LuaMethodOverload(
+                    argumentTypes = String.class,
+                    argumentNames = "taskName"
+            ),
+            value = "model_part.new_block")
```
4P5: Was `model_part.new_block` accidentally left in?

UnlikePaladin:
lgtm

PR #174, opened 2024-02-07
Texture:applyMatrix() clipping

4P5:
This PR adds a new field to Texture:applyMatrix(), `clip`, which clamps the resulting colours' channels between 0 and 1.


4P5:
firefox_3x6QLXdciP

PR #185, opened 2024-02-19
Add an event for modifying sign text

PoolloverNathan:
It takes the sign position (which allows deriving the rotation and type via `WorldAPI:getBlockState()`, a mutable array of lines, and whether it is the front side. This event is also only fired if chat messages are enabled.

Untested on Forge.

(cherry picked from commit 09cbc72ecd5108b52b91cfa348ff512ba10b671b)

---

This PR is intended to be merged via rebase or local fast-forward.


applejuiceyy:
is adding 1 single event for 1 single block a good thing whilst balancing out the risks of feature creeping?

UnlikePaladin:
This is super out of scope imo, check discord for more info, it doesn't have a place in Figura

PR #187, opened 2024-02-22
Retry saving config files

PoolloverNathan:
Contains a questionable Thread.sleep but it's capped to 25ms and should otherwise be fine. Also exposes the error message to the user.

### DIFF
--- a/common/src/main/java/org/figuramc/figura/lua/api/ConfigAPI.java
+++ b/common/src/main/java/org/figuramc/figura/lua/api/ConfigAPI.java
@@ -79,7 +79,7 @@ public class ConfigAPI {
         }
     }
 
-    private void writeWithRetries(Path path, byte[] data, int retries) {
+    private void writeWithRetries(Path path, byte[] data, int retries) throws Exception {
         try (OutputStream fs = Files.newOutputStream(path)) {
 
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/figuramc/figura/lua/api/ConfigAPI.java
```diff
@@ -79,6 +79,19 @@ private Path getPath() {
         }
     }
 
+    private void writeWithRetries(Path path, byte[] data, int retries) {
```
PoolloverNathan: I'm here to play catch, not write a 30-page paper on the intricacies of the game we know as 'catch'.

```suggestion
    private void writeWithRetries(Path path, byte[] data, int retries) throws Exception {
```

common/src/main/java/org/figuramc/figura/lua/api/ConfigAPI.java
```diff
@@ -79,6 +79,19 @@ private Path getPath() {
         }
     }
 
+    private void writeWithRetries(Path path, byte[] data, int retries) throws Exception {
+        try (OutputStream fs = Files.newOutputStream(path)) {
+
```
PoolloverNathan:
```suggestion
            fs.write(data);    
```

PoolloverNathan (commented):
you idiot

PR #200, opened 2024-03-08
Don't parse metadata into json 3 seperate times and Client.isIntegratedServer

superpowers04:
accidentally pushed client.isIntegratedServer to here so i guess it lives rent free in this pr now


superpowers04:
the error was due to out of memory :skull:

PoolloverNathan:
> the error was due to out of memory 💀

dementia

PR #223, opened 2024-05-23
Removed random green pixels

ToastTextures:
ARE THEY HERE
WHO LET THIS HAPPEN
WHO APPROVED THIS
HOW DO YOU GET GREEN????
THIS WASN'T PART OF THE CERTIFIED FIGURAMC PALETTE?
ARE THESE HERE ON PURPOSE???
DO THEY SERVE A PURPOSE?
I am in pain
Also what's with the gradient in the figura icon????
Also like why is there a random dark red color here instead of the black on the github and kofi logos
Who approved this
So unacceptable
Can never unsee it
:sob:


aprilkenopsia:
im zorry..

aprilkenopsia:
these are being reworked btw we just havent implemented new stuff yet

UnlikePaladin:
real

PR #225, opened 2024-05-25
fix getShaderPackName

soup587:
also updates the iris version and gradle this probably sucks

is also my first pull request ever, any advice is accepted !!


soup587:
supposedly can be reworked to work with multiple iris versions

UnlikePaladin:
I fixed it myself, dw but thanks

PR #239, opened 2024-07-16
Figura Server side (or FSB)

lexize:
Main goal of this PR is make Figura even more extendable, make it so main backend is not dying from a flood of a new Figura users, and also make it easier for server to communicate with avatars.

What exactly this PR is going to add:
1. **Server side of Figura.** Useful for server owners, and safe enough for players. All the avatars' hashes will be checked both on server and then on client side, in order to be sure that server is sending exactly the avatar that has been requested. Host's avatars will be checked with ehashes (will explain what it is later), in order to check that host won't get unknown avatars equipped.
2. **EHashes and secret key.** E-Hash stands for "encrypted hash" (yes i know im very original and i mean this is a concept that will be used only inside project and players won't even need to care about it). It is used in order to check avatar ownership on host side. EHash is a hash-sum of avatar's hash, and hash of a secret key set in Figura's settings. In addition to basic hash checking client will check if incoming avatar ehash sent by a server equals to expected ehash. If it does, avatar will be allowed to be equipped, otherwise avatar won't be equipped and client will be notified about possible attempt of avatar replacement. P.S. In order to make uploaded avatar be allowed on multiple clients (for example, multiple instances or devices) client will have to set the same secret key in Figura's settings.
3. **Server events.** Like Bukkit events, used to extend server side of Figura yet still keep it safe enough. It can be used to provide a different source of avatars (for example, DB), allow or disallow client connection, and sending/receiving figura packets to/from client.
4. **Figura packets** A way for a Figura server and client sides to communicate. This can be used if your server implements something that could be listened and used within your Figura avatar, for example a global server event, or something like that. (This is advanced feature that uses FiguraBuffer as input or output).

Changes so far:
* Modified root gradle project and `gradle.properties` in order to apply mappings and include minecraft only in projects that are modded.
* Started writing server side of Figura, added bunch of packet classes.
* Added interface for platform independent FriendlyByteBuf, so all the needed functions can be also used in Bukkit plugin
* Added platform independent Identifier record, which is basically ResourceLocation, but both for Bukkit and Fabric/Forge.
* Started writing AvatarManager for server side.
* Added a little Events system to make figura server side extendable but also secure enough. There are several types of events: Regular ones, cancellable, and returnable. Returnable events can be used
for custom avatar/metadata/usedata loading implementation, for example from DB (by default everything including avatars, avatar metadata, and userdata, will be stored in a folder)

For now I'm gonna take a break from working on this project, so i am just making it a pull request to make it possible for maintainers to also make changes if they wish.


PoolloverNathan:
Would it be possible to use Minecraft's existing chat signing key to sign avatars? This could be used to avoid needing to share the key manually, while also preventing other players (or the server, if it has the key) replacing avatars.

lexize:
Key won't be shared at all. It will be only used to check ownership of host avatars, and will be accessible only from host client. Sharing it means allowing a server to just replace your avatar without your permission. Also I suggest to just go to thread on discord cuz we are basically discussing everything FSB related in there.

About using this chat signature thing - I don't even know how it works internally.

TheBunnyMan123:
> Would it be possible to use Minecraft's existing chat signing key to sign avatars? This could be used to avoid needing to share the key manually, while also preventing other players (or the server, if it has the key) replacing avatars.

what if someone (like me) doesn't sign chat messages using something like no chat reports

UnlikePaladin:
the time has come

ElectricDaFearox:
"it's time to go to heaven my child"

UnlikePaladin (approved):
chonker PR, thanks Kate!

PR #245, opened 2024-07-25
Include port for backend SERVER_IP

Leialoha:
I've been working on developing the backend for Figura clients. For additional context, please refer to issue #239.

The backend setup is nearly complete, but I'm encountering a problem where the client doesn't retain the specified port. While this may not seem critical, it becomes an issue when using a server hosting service. Without proper port forwarding, which must be handled by the server host, the client is unable to access the necessary port. Specifically, the client requires port 443 to function, and currently, the mod/plugin is restricted to using only this port. This also necessitates running the application with administrative permissions.

This pull request addresses the port retention issue, allowing the mod/plugin to be used on ports other than 443. This fix will enable me to continue developing and testing the backend in a production environment.

If you would like to review the code, please refer to my repository: [https://github.com/Leialoha/Figura-Backend-Server](https://github.com/Leialoha/Figura-Backend-Server).

### DIFF
--- a/common/src/main/java/org/figuramc/figura/backend2/HttpAPI.java
+++ b/common/src/main/java/org/figuramc/figura/backend2/HttpAPI.java
@@ -29,8 +29,18 @@ public class HttpAPI {
     }
 
     protected static String getBackendAddress() {
+        return "https://" + getBackendAddressWithPort() + "/api";
+    }
+
+    private static String getBackendAddressWithPort() {
         ServerAddress backendIP = ServerAddress.parseString(Configs.SERVER_IP.value);
-        return "https://" + backendIP.getHost() + "/api";
+        boolean hasPort = Configs.SERVER_IP.value.matches("[^:]+:\\d+");
+        if (hasPort) {
+            try {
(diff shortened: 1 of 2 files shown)


UnlikePaladin:
While we as a team are not particularly happy with the existence of third party backends because of the risks they may entail for users, which is why FSB is being developed in house as it is safer and allows servers to function as their own backend anyway, this doesn't really change anything that significant or make something that wasn't possible before a possibility so I guess we can accept this.

Leialoha:
Ah, my goal wasn't to have Figura Backend 3.0 but was to create a plugin for Figura to run on Bukkit servers. Can't wait to see the implementation you guys create and I honestly didn't know you were actively working on it.

lexize:
FSB will aslo be available on Bukkit. I am trying to make it as platform independent as possible. So far planned platforms are Fabric, Forge, and Bukkit. List of platforms can get larger in future.
Also we are trying to make it as extendable as possible, but yet keep it safe.
If you want to learn more we have a thread in FiguraMC: https://discord.com/channels/1129805506354085959/1262676821338493019

PR #258, opened 2024-09-10
removed nil behavior with setVisible

KitCat962:
Suggestion from JimmyHelp in [discord](https://discord.com/channels/1129805506354085959/1282010060897128510)

Removes all traces of "nil means apply from parent" from the `setVisible` function, as it is impossible for a child to not follow parent's visibility due to a previous update.
`nil` now means `false`, which is expected behaviour for value coercion.


Riftlight:
Typo in the lang file new desc: "Works recursivly" :grin:

PR #273, opened 2024-11-17
unfeat: remove default permissions

PoolloverNathan:
removes the old backend permission system left behind by F, which doesn't respect user config


TheBunnyMan123:
This had reason to exist before, but I agree. It should either be removed or have an option to be toggled off

PoolloverNathan:
@vickystxr authorship given as requested

envicious:
rizz ohio

PR #280, opened 2024-12-14
Update CI to v4, Do not persist credentials

Kichura:
Mainly to comply with [GitHub's breaking changes blog](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions) and also because `zizmor` spotted credentials being persistent which is unsafe, Ubuntu release on other hand can stay frozen mainly due to stability reasons.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


UnlikePaladin:
thanks!

UnlikePaladin:
wait no, they were targeting the wrong branch darnit, always target 1.20, not 1.21.3

PR #285, opened 2024-12-18
fix non-looped animation interpolation

soup587:
(almost 2am commit, please tell me if i'm doing any of these poorly)


PoolloverNathan (approved):
LGTM (I have not tested or looked at the code whatsoever)

soup587:
at the very least it does in fact fix non-looped animation interpolation for the ones i tested, and doesn't seem to break others from my testing

TheBunnyMan123 (approved):
Yes, this is in fact code

soup587:
just realised i think this actually broke looped interpolation

soup587:
(it's not as noticeable as non-looped interpolation but like  it should not breaked it)

soup587:
nvm it's not a regression. may still look into it later

PR #289, opened 2024-12-21
feat: host event for detecting/cancelling particles

PoolloverNathan:


TheBunnyMan123:
Since this is a fix that means it'll get merged asap right? :trol:

PR #290, opened 2024-12-23
Update PlayerRendererMixin.java

Riftlight:
Fix the player nameplate being changed while dead causing a crash due to return value of getEntity being null


Riftlight:
i think github actions is high

Riftlight:
pretend this comment never happened (not deleting because uhhhhh i don’t feel like it)

Riftlight:
jesus. turns out everything is fine and i mixed up the reds and greens of the diff. kill me

PR #295, opened 2025-01-15
Fix vertices being sent to the GPU that are being changed when they shouldn't be maybe?

superpowers04:
^

### DIFF
--- a/common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
+++ b/common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
@@ -17,6 +17,7 @@ import org.figuramc.figura.config.Configs;
 import org.figuramc.figura.lua.api.ClientAPI;
 import org.figuramc.figura.math.matrix.FiguraMat3;
 import org.figuramc.figura.math.matrix.FiguraMat4;
+import org.figuramc.figura.math.vector.FiguraVec2;
 import org.figuramc.figura.math.vector.FiguraVec3;
 import org.figuramc.figura.math.vector.FiguraVec4;
 import org.figuramc.figura.model.*;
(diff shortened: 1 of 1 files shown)


penguinencounter:
why is every line in that file marked as changed

edit: they converted it to tabs

PoolloverNathan:
> why is every line in that file marked as changed

whitespace changes, presumably? looks somewhat sane when I disable whitespace

PoolloverNathan:
> why is every line in that file marked as changed
> 
> edit: they converted it to tabs

oh god

superpowers04:
This was just a copy-paste from Extura tbh, that's why the file is formatted weirdly, it follows Extura's formatting

UnlikePaladin:
too vague, explain the fix further and list your changes or fix your formatting to match Figura's, I'm not going to try and diff that :ioa:

penguinencounter:
1. https://discord.com/channels/1129805506354085959/1129811275380162730/1328613173183385641
2. https://discord.com/channels/1129805506354085959/1129811275380162730/1328926184171503657

@superpowers04: I'm a bit concerned about the fact that this does `N*(vertex count)` allocations every frame for a feature that won't be used in 99.9% of cases. In the original, there's no allocation overhead because it keeps re-using the same vector reference. Maybe a way to only do the copying if necessary would be better?

penguinencounter:
also issue tldr: in this very specific set of circumstances:

1. a part with multiple parents (or is being rendered multiple times for any reason)
2. with modified vertices between each rendering (i.e. by hooking `preRender` or `midRender` on each parent and changing the vertex data)

... the actually rendered vertices on all instances of the part being rendered are the last values before the frame ends. This doesn't really make sense if you're expecting the vertex data to be committed when the part is rendered, but it's really deferred into the buffer queue thing, which uses a reference to the part's vertex data instead of copying it when `pushToBuffer` is called

superpowers04:
Closed because I am a dumb, made the commit wrong and added a lot of overhead on accident. Will fix later

PR #296, opened 2025-01-16
Send copies of vertexes instead of pointers to changing vertices pt 2 electric boogaoo + some other rendering stuff maybe?

superpowers04:
pull/295 but less bad maybe

### DIFF
--- a/common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
+++ b/common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
@@ -17,6 +17,7 @@ import org.figuramc.figura.config.Configs;
 import org.figuramc.figura.lua.api.ClientAPI;
 import org.figuramc.figura.math.matrix.FiguraMat3;
 import org.figuramc.figura.math.matrix.FiguraMat4;
+import org.figuramc.figura.math.vector.FiguraVec2;
 import org.figuramc.figura.math.vector.FiguraVec3;
 import org.figuramc.figura.math.vector.FiguraVec4;
 import org.figuramc.figura.model.*;
@@ -104,7 +105,7 @@ public class ImmediateAvatarRenderer extends AvatarRenderer {
 
         // Set shouldRenderPivots
         int config = Configs.RENDER_DEBUG_PARTS_PIVOT.value;
-        if (!Minecraft.getInstance().getEntityRenderDispatcher().shouldRenderHitBoxes() || (!avatar.isHost && config < 2))
+        if ((!avatar.isHost && config < 2) || !Minecraft.getInstance().getEntityRenderDispatcher().shouldRenderHitBoxes())
             shouldRenderPivots = 0;
         else
             shouldRenderPivots = config;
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
```diff
@@ -542,32 +529,81 @@ private void pushToBuffer(int faceCount, VertexData vertexData, PartCustomizatio
...
     }
 
+    private static class ToBeConsumedVertexData {
+        public Float x,y,z;
+        public Float r,g,b,a;
+        public Float uvX,uvY;
+        public int overlay;
+        public int light;
+        public Float nX,nY,nZ;
+        public boolean consumed;
```
penguinencounter, line 570: any reason these are capital `Float` (the box type) instead of lowercase `float` (the primitive)? none of them look nullable to me

common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
```diff
@@ -181,8 +180,7 @@ protected PartCustomization setupRootCustomization(double vertOffset) {
         customization.setPrimaryRenderType(RenderTypes.TRANSLUCENT);
         customization.setSecondaryRenderType(RenderTypes.EMISSIVE);
 
-        double s = 1.0 / 16;
-        customization.positionMatrix.scale(s, s, s);
+        customization.positionMatrix.scale(0.0625, 0.0625, 0.0625);
```
UnlikePaladin, line 183: Either add a comment explaining why the 0.0625, that is clarify it stands for 1 / 16, or undo the change. The compiler or jvm probably inline this anyway because it's a constant so it's an unnecessary change that makes it less clear and makes changes in the future harder because no one will know what 0.062 is or stands for.

common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
```diff
@@ -268,29 +266,17 @@ protected boolean renderPart(FiguraModelPart part, int[] remainingComplexity, bo
...
                 updateLight = false;
-                pivotOffsetter.light = custom.light;
-            }
             else if (updateLight && (l = Minecraft.getInstance().level) != null) {
-                FiguraVec3 pos = part.savedPartToWorldMat.apply(0d, 0d, 0d);
-                int block = l.getBrightness(LightLayer.BLOCK, pos.asBlockPos());
-                int sky = l.getBrightness(LightLayer.SKY, pos.asBlockPos());
-                customizationStack.peek().light = LightTexture.pack(block, sky);
+                var pos = part.savedPartToWorldMat.apply(0d, 0d, 0d).asBlockPos();
+                customizationStack.peek().light = LightTexture.pack(l.getBrightness(LightLayer.BLOCK, pos), l.getBrightness(LightLayer.SKY, pos));
```
UnlikePaladin, line 278: Don't use things such as var, this is not JS and I have to back port all these changes to Java 8 later, there's no good reason to. Also why inline these? They're locals and get destroyed after they go out of scope anyway, it's not optimizing anything

common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
```diff
@@ -542,32 +529,81 @@ private void pushToBuffer(int faceCount, VertexData vertexData, PartCustomizatio
 
         int overlay = customization.overlay;
         int light = vertexData.fullBright ? LightTexture.FULL_BRIGHT : customization.light;
+        ToBeConsumedVertexData[] vertexDatas = new ToBeConsumedVertexData[vertCount]; // why the fuck did i use duckai for this
```
UnlikePaladin, line 532: Please provide a visual example of what the change is actually fixing or doing, I understand you're "copying" the data to consume it later but it is less than clear why you are doing so. Explain why the vertices change, etc.

UnlikePaladin (changes requested):
Please read the observations

common/src/main/java/org/figuramc/figura/model/rendering/ImmediateAvatarRenderer.java
superpowers04: Typo, Used to other languages using Float instead of float

superpowers04:
This doesn't actually fix the issue at the moment, the requested changes have been made but I don't have enough time at the moment to fix the issue :c_:, I'll push the changes when it's done

PR #306, opened 2025-02-04
Add getCameraMatrix method to ClientAPI

Evoloxi:
Requested by GNamimates, saves a bunch of unnecessary calculation


UnlikePaladin (commented):
the implementation looks fine, but I do not like this and would personally not rely on it because the matrix changes between versions at Mojang's will

PR #314, opened 2025-02-18
Add promise-like callbacks to futures

PoolloverNathan:
This pull request adds several utility methods to `Future`s:

* `onFinish` — calls the given function when the future resolves successfully, and the second optional function when it errors (like JS `Promise#then`)
* `onFinishError` — calls the given function when the future errors (like JS `Promise#catch`)
* `hasError` — pre-existing method, but actually exposed to Lua since I guess it wasn't (idk I'm looking at diffs after-the-fact)
* `map` — returns a new future that succeeds when the original future succeeds, modifying the value with the given function; passes errors (like `then`)
* `andThen` — like `map`, but the function returns another future, and the returned future settles when the returned future settles or the outer fails
* `handle` — returns a future that settles when the original future settles, but handles errors using the given function

and some utilities:
* `Avatar#submit` to add a function to the event queue
* futures now track their avatars
* renamed `Future#handle` to `complete` to avoid overloads
* `LuaCloseable` to provide one `close` method to Lua
* better toString for futures
* `net.newFuture`, returns `fut, complete, error` and acts like JS `new Promise` if given a function argument

# Tests
### `onFinish`
- [ ] calls the first function on success
- [ ] calls the second function on error
- [ ] has no behavior on error if no second function is given
- [ ] does not call the first function on error
- [ ] does not call the second function on success
- [ ] errors if no first function is given
- [ ] propagates an error in the first callback to the returned future
- [ ] propagates an error in the second callback to the returned future
### `onFinishError`
- [ ] errors if no function is given
- [ ] calls the given function on error
- [ ] does not call the given function on success
- [ ] propagates an error in the callback to the returned future
### hasError
- [ ] returns false before the future is settled
- [ ] returns false when the future is settled with success
- [ ] returns true when the future is seddled with an error
- [ ] propagates an error in the callback to the returned future
### `map`
- [ ] calls the function on success
- [ ] immediately succeeds the returned future when the future succeeds
- [ ] does not call the function on error
- [ ] errors the returned future immediately when the future errors
- [ ] propagates an error in the callback to the returned future
### `andThen`
- [ ] calls the function on success
- [ ] immediately succeeds the returned future when the callback's returned future succeeds
- [ ] does not call the function on error
- [ ] errors the returned future immediately when the receiver future errors
- [ ] errors the returned future immediately when the callback's future errors
- [ ] propagates an error in the callback to the returned future
### `handle`
- [ ] calls the function on error
- [ ] does not call the function on success
- [ ] propagates errors to the returned future
- [ ] propagates an error in the callback to the returned future


envicious:
Bug? error? glitch?

image
Just doesn't work

how to repro

```lua
function Test()
    local fut = net.http:request("https://icanhazdadjoke.com"):method("GET"):header("accept","text/plain"):send()
    print(fut:onFinish(
        function(...)
            print(...)
        end,
        function(...)
            print(...)
        end));
end

Test()
```

PR #328, opened 2025-03-01
Fix addScript not removing scripts and missing docs

auriafoxgirl:
Fixes `addScript(path, nil)` not removing script from nbt
and also fixes docs in `/figura docs global addScript/getScript/getScripts`

Meow

### DIFF
--- a/common/src/main/java/org/figuramc/figura/lua/docs/FiguraGlobalsDocs.java
+++ b/common/src/main/java/org/figuramc/figura/lua/docs/FiguraGlobalsDocs.java
@@ -112,6 +112,35 @@ public abstract class FiguraGlobalsDocs {
 
     @LuaMethodDoc(
             overloads = {
+                    @LuaMethodOverload(
+                            argumentTypes = String.class,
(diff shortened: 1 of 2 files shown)


Evoloxi:
Meow (LGTM)

common/src/main/java/org/figuramc/figura/lua/docs/FiguraGlobalsDocs.java
```diff
@@ -110,6 +110,35 @@ public abstract class FiguraGlobalsDocs {
...
+    public static Object getScripts() {return null;}
+
+    @LuaMethodDoc(
+            overloads = @LuaMethodOverload(
+                    argumentTypes = String.class,
+                    argumentNames = "scriptName"
+            ),
+            value = "globals.get_script"
+    )
+    public static Object getScript() {return null;}
```
Evoloxi, line 140: Except for one thing: 
I know the other methods are also formatted like this but it just feels... wrong.

```suggestion
    public static Object getScript() {
        return null;
    }
```

PR #332, opened 2025-03-04
Rewrite the action wheel to use model parts

UnlikePaladin:
The player is now able to use model parts including groups on the actions as well as items and sprites/textures for legacy compatibility


UnlikePaladin:
wait tf, did I never finish this

UnlikePaladin:
okay finished now

Bitslayn:
Is this creating an ItemTask for Action Wheel items? ItemTasks don't work for player heads if that head is rendered by the skull owner (intended behavior) causing the skull to appear as a vanilla head; if that is what causes #451

UnlikePaladin:
yeah, I just noticed that 🫠

Bitslayn:
Maybe you could use your rendering magic to make ItemTask skulls work only in the action wheel?

Bitslayn:
While also avoiding an infinite render loop

PR #339, opened 2025-03-14
Add avatar:getPermissionValue() to get the permission level as an int instead of a string.

MEEPofFaith:


MEEPofFaith:
image

PoolloverNathan (changes requested):
I feel like this would encourage checking the permission level directly, when it'd be more accurate to check the actually-needed permissions. Several people on our server have a customized DEFAULT level with higher permissions.

PR #346, opened 2025-03-29
Nameplate unload crash fix (1.21.3 edition)

penguinencounter:
Fix the player nameplate being changed while dead causing a crash due to return value of getEntity being null

so it turns out the changes were actually on the 1.21.3 version and they were just... missing? this is already on 1.21.4 from #290

### DIFF
--- a/common/src/main/java/org/figuramc/figura/mixin/render/renderers/PlayerRendererMixin.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/render/renderers/PlayerRendererMixin.java
@@ -112,7 +112,9 @@ public abstract class PlayerRendererMixin extends LivingEntityRenderer<AbstractC
 
         // badges
         FiguraMod.popPushProfiler("badges");
-        replacement = Badges.appendBadges(replacement, Minecraft.getInstance().level.getEntity(player.id).getUUID(), config > 1);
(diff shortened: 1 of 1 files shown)


Riftlight:
thank you :pray: i blame the fact that nobody uses 1.21.3

common/src/main/java/org/figuramc/figura/mixin/render/renderers/PlayerRendererMixin.java
```diff
@@ -112,7 +112,9 @@ private Component modifyPlayerNameText(Component text, @Local(argsOnly = true) P
 
         // badges
         FiguraMod.popPushProfiler("badges");
-        replacement = Badges.appendBadges(replacement, Minecraft.getInstance().level.getEntity(player.id).getUUID(), config > 1);
+        if (Minecraft.getInstance().level.getEntity(player.id) != null) { // null while dead
+			replacement = Badges.appendBadges(replacement, Minecraft.getInstance().level.getEntity(player.id).getUUID(), config > 1);
+		}
```
UnlikePaladin: pls fix the indentation? ty 😭

penguinencounter: done

UnlikePaladin (approved):
damn that was quick

PR #352, opened 2025-04-07
Improve Error Handling in AvatarRenderer for Missing Texture Data

VenRoot:
This PR addresses an issue in the `AvatarRenderer` class where missing byte array data for texture keys could lead to potential errors. The changes introduce improved error handling by logging warnings when texture data is missing and ensuring the program continues to run smoothly without crashing.

### Changes:
* Added import for `FiguraMod`.
* Updated the AvatarRenderer constructor to use `Optional.orElseGet()` for handling missing byte array data.
* Log a warning message when byte array data for a texture key is missing.
* Ensured that an empty byte array is used as a fallback when data is missing.

## Testing:

* Ensured that the application does not crash, correctly displays the model and continues to function correctly with the fallback empty byte array.

* Ensured that the warning is logged correctly

### DIFF
--- a/common/src/main/java/org/figuramc/figura/mixin/render/TextureAtlasAccessor.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/render/TextureAtlasAccessor.java
@@ -10,7 +10,7 @@ import org.spongepowered.asm.mixin.gen.Invoker;
 
 import java.util.Map;
 
-@Mixin(TextureAtlas.class)
+@Mixin(value = TextureAtlas.class, priority = 900)
 public interface TextureAtlasAccessor {
     @Intrinsic
     @Accessor("texturesByName")
--- a/common/src/main/java/org/figuramc/figura/mixin/render/TextureAtlasAccessor.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/render/TextureAtlasAccessor.java
@@ -18,9 +18,9 @@ public interface TextureAtlasAccessor {
 
     @Intrinsic
     @Invoker("getWidth")
-    int getWidth();
+    int figuraGetWidth();
 
     @Intrinsic
     @Invoker("getHeight")
-    int getHeight();
+    int figuraGetHeight();
 }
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/figuramc/figura/model/rendering/AvatarRenderer.java
```diff
@@ -87,7 +88,11 @@ public AvatarRenderer(Avatar avatar) {
 
         // src files
         for (String key : src.keySet()) {
-            byte[] bytes = src.getByteArray(key).get();
+            byte[] bytes = src.getByteArray(key).orElseGet(() -> {
+                FiguraMod.LOGGER.warn("Missing byte array data for texture key: " + key);
```
UnlikePaladin, line 92: I don't think we need a warning here, it would confuse users, plus whenever something went horribly wrong it would probably spam the log quite badly. Vanilla already used to return an empty byte array with no warning before they changed it to optionals pre 1.21.5 and that is the behavior that should be replicated.

common/src/main/java/org/figuramc/figura/mixin/render/TextureAtlasAccessor.java
```diff
@@ -10,17 +10,17 @@
...
+@Mixin(value = TextureAtlas.class, priority = 900)
 public interface TextureAtlasAccessor {
     @Intrinsic
     @Accessor("texturesByName")
     Map<ResourceLocation, TextureAtlasSprite> getTexturesByName();
 
     @Intrinsic
     @Invoker("getWidth")
-    int getWidth();
+    int figuraGetWidth();
```
UnlikePaladin, line 21: The standard in the codebase and mixin is to modID$identifier so figura$getWidth, please stick to the conventions.

UnlikePaladin (commented):
other than that the changes/fixes are appreciated

UnlikePaladin:
also you literally added the sodium fix someone had already pred into 1.20 here, plus that changelog looks ai generated as hell 😭  no offense

PR #357, opened 2025-04-25
Transition to ARM, JDK 21.0.7, Gradle 8.14

Kichura:
GitHub can now happily make use of the ARM64 architecture which is safe to transition to in terms of this repo,
As for the JDK - it has been bumped to 21.0.7 and Gradle has been bumped to 8.14.


penguinencounter:
why do we need this? it's not like we're compiling natives or anything, the platform architecture shouldn't matter

Kichura:
> why do we need this? it's not like we're compiling natives or anything, the platform architecture shouldn't matter

It's mainly for efficiency rather than compiling natives as java is universal.

PoolloverNathan:
> It's mainly for efficiency rather than compiling natives as java is universal.

Yes, but what does using ARM gain us over current runners?

TheBunnyMan123:
> > It's mainly for efficiency rather than compiling natives as java is universal.
> 
> Yes, but what does using ARM gain us over current runners?

We save GitHub a few cents of power

I vote we make it less efficient to compile

PR #358, opened 2025-05-01
FSB: fix crashes, and a few UX things

penguinencounter:
* fix crash caused by race condition between network and server threads when uploading an avatar
* put error handling pretty much everywhere so errors don't take down the entire server
* new packet types for server to reply with when avatar uploads and deletions finish processing
  * uploading to Server Cloud now waits for the server to finish the upload and reply with `S2CAvatarReadyPacket` before equipping the avatar and displaying the upload toast
  * uploading to backend only now doesn't try to equip the avatar on FSB if it's connected
  * still tries to use the FSB avatar when uploading to both
  * client now displays the Avatar Deleted toast!
* corrupt avatar files (blank / missing) should now be handled much better (at least they don't crash now?)


UnlikePaladin (approved):
lgtm overall even if I'm not a fan of making loadAvatar throw an IO Exception, guess it's better than a runtime exception though

PR #363, opened 2025-05-02
Custom error message about forgetting colons

penguinencounter:
Specifics of error message wording are still up for debate.

Hopefully this doesn't break existing code. (In a perfect world we wouldn't be using error message hacks but we don't live in that world)


UnlikePaladin (approved):
mfw figura has decent error messages

PR #364, opened 2025-05-06
Figura icons & names in SVC

KingRoboLizard:
Adds compat between Figura and SVC to make SVC render the figura icons and names instead of the vanilla ones.


envicious:
Good implementation, SVC is not required, works just like the tab list, and has scrolling for longer names.

envicious (approved):

UnlikePaladin:
we literally already had this but I removed it because it frequently broke with changes on SVC's side so this is probably better off as a small plugin or add-on.

Bitslayn:
> we literally already had this but I removed it because it frequently broke with changes on SVC's side so this is probably better off as a small plugin or add-on.

😢

PR #369, opened 2025-05-24
Fix load() environment issue

penguinencounter:
should keep the actual environment replacement functionality working, but also keep tracebacks etc intact


Bitslayn:
What's this do? Just add the environment string automatically so you don't have to define it for things like loading scripts from the data folder?

Fnige:
> What's this do? Just add the environment string automatically so you don't have to define it for things like loading scripts from the data folder?

I _think_ it might be related to this conversation? https://discord.com/channels/1129805506354085959/1129815148190761081/1375702674162585640

PR #374, opened 2025-06-15
Add side argument to addScript()

Captain8771:
this change allows for not adding the added script to the NBT of the avatar that gets uploaded, while still adding to the runtime, and vice versa. this change is backwards compatible with prior avatars that use addScript as-is (aka the default side is "BOTH")


UnlikePaladin:
Sounds cursed but sure ig

PR #377, opened 2025-06-21
Fine-grained control of Rate, size and count limits using meta permissions in FSB

dimaguy:
And unify permission nodes under its own enum for consistency

### DIFF
--- a/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
+++ b/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
@@ -25,4 +25,13 @@ public enum FiguraPermissionNodes {
     public String toString() {
         return node;
     }
+
+    public static FiguraPermissionNodes fromString(String node) {
+        for (FiguraPermissionNodes permissionNode : values()) {
+            if (permissionNode.node.equalsIgnoreCase(node)) {
+                return permissionNode;
+            }
+        }
+        throw new IllegalArgumentException("No permission node found for: " + node);
+    }
 }
--- a/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
+++ b/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
@@ -11,7 +11,7 @@ public enum FiguraPermissionNodes {
 
     FIGURA_PINGS_SIZELIMIT("figura.pings.sizelimit"),
 
-    FIGURA_AVATARS_SIZELIMIT("figura.avatars.send"),
+    FIGURA_AVATARS_SIZELIMIT("figura.avatars.sizelimit"),
 
     FIGURA_AVATARS_COUNTLIMIT("figura.avatars.countlimit");
 
--- a/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
+++ b/server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
@@ -1,0 +1,28 @@
+package org.figuramc.figura.server;
+
+public enum FiguraPermissionNodes {
+    FIGURA_AVATARS_IMMORTALIZE("figura.avatars.immortalize"),
+
+    FIGURA_AVATARS_SET("figura.avatars.set"),
+
+    FIGURA_AVATARS_CLEAR("figura.avatars.clear"),
+
+    FIGURA_PINGS_RATELIMIT("figura.pings.ratelimit"),
+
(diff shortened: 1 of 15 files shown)


lycanea:
this is such a perfectly crafted creation that should be merged IMMEDIATELY, this is a work of art that shall not be neglected, this is the best pull request I've seen and likely will see

PoolloverNathan (approved):
We're using Fabric API's permissions, correct? The only problem I could see is a hard dependency on Fabric API, but if someone calls Fabric API a problem I'm revoking their mixin privileges for a week.

lycanea (approved):

dimaguy:
> We're using Fabric API's permissions, correct? The only problem I could see is a hard dependency on Fabric API, but if someone calls Fabric API a problem I'm revoking their mixin privileges for a week.

Uhhh the dependency was already there, there is some level of permission support on figura already, I just augmented the reliance on it

PoolloverNathan:
> > We're using Fabric API's permissions, correct? The only problem I could see is a hard dependency on Fabric API, but if someone calls Fabric API a problem I'm revoking their mixin privileges for a week.
> 
> Uhhh the dependency was already there, there is some level of permission support on figura already, I just augmented the reliance on it

oh! then yeah, LGTM (I have no power here)

penguinencounter:
/genq: why are there so many conversions between integers and strings?

dimaguy:
> /genq: why are there so many conversions between integers and strings?

The common denominator of meta permissions across the three implementations of servers seemed to be the string value, yet the end result is an integer, this could be perhaps simplified to some degree, but not by much

server-common/src/main/java/org/figuramc/figura/server/FiguraPermissions.java
```diff
@@ -1,12 +1,19 @@
...
 
 import java.util.List;
 
 public class FiguraPermissions {
-    public static final List<Pair<String, Boolean>> PERMISSIONS_LIST = List.of(
-            Pair.of("figura.avatars.immortalize", false),
-            Pair.of("figura.avatars.set", false)
+    public static final List<Pair<FiguraPermissionNodes, Boolean>> PERMISSIONS_LIST = List.of(
+            Pair.of(FIGURA_AVATARS_IMMORTALIZE, false),
+            Pair.of(FIGURA_AVATARS_SET, false)
```
penguinencounter, line 11: could maybe put `FIGURA_AVATARS_CLEAR` in here while you're at it, it's currently inaccessible in MinecraftForge

server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
```diff
@@ -0,0 +1,37 @@
...
+
+    FIGURA_AVATARS_SET("figura.avatars.set"),
+
+    FIGURA_AVATARS_CLEAR("figura.avatars.clear"),
+
+    FIGURA_PINGS_RATELIMIT("figura.pings.ratelimit"),
+
+    FIGURA_PINGS_SIZELIMIT("figura.pings.sizelimit"),
+
+    FIGURA_AVATARS_SIZELIMIT("figura.avatars.send"),
```
penguinencounter: why is this called `send`?

penguinencounter (approved):
code looks OK overall, works when tested on Fabric/Forge

a neat feature (not required) would be to somehow update the information display in-game but there are currently no packets that let you do that without killing the connection

(I still don't have any power here)

server-common/src/main/java/org/figuramc/figura/server/FiguraPermissionNodes.java
dimaguy: Smol distraction, Ill fix when I get to computer

server-common/src/main/java/org/figuramc/figura/server/FiguraPermissions.java
dimaguy: I could yeah, I just didn't do it yet because I didnt know why clear wasnt there before I got to touch the code

dimaguy:
> a neat feature (not required) would be to somehow update the information display in-game but there are currently no packets that let you do that without killing the connection

I wonder what would happen if I just resent the server handshake packet, but the thing is, there's no event for the permissions changing at the moment, best alternative would be a keepalive packet sent regularly with the up to date permissions

UnlikePaladin (approved):
Very nicely crafted, looks good to me! Many thanks for the pr!

dimaguy:
> Very nicely crafted, looks good to me! Many thanks for the pr!

Glad you like it! Should I add the clear permission node to the forge list as well? I wasn't sure if it was an oversight, so I left as-is

PR #378, opened 2025-06-25
Opus: Don't crash if passed a really tiny (or limited) buffer

penguinencounter:
TACZ is at it again, apparently: https://discord.com/channels/1129805506354085959/1129811275380162730/1387215984674209974

### DIFF
--- a/common/src/main/java/org/figuramc/figura/mixin/sound/OggAudioStreamMixin.java
+++ b/common/src/main/java/org/figuramc/figura/mixin/sound/OggAudioStreamMixin.java
@@ -47,6 +47,8 @@ public abstract class OggAudioStreamMixin {
         }
 
         byte[] headerBytes = new byte[8];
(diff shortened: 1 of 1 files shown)


Evoloxi:
Why the hell would anyone pass in a buffer that small?



An ogg vorbis packet doesn't even contain any audio data until the 28th byte

penguinencounter:
@Evoloxi my only guess is that there's a corrupt audio file in the TACZ resources or the way they're reading it is somehow wrong [(with `Resource::open`)](<https://github.com/MCModderAnchor/TACZ/blob/1.20.1/src/main/java/com/tacz/guns/client/resource/manager/SoundAssetsManager.java#L40>)

I figured that if the audio file is that messed up we just let Minecraft's loader handle it

UnlikePaladin:
Bruh but thanks for the fix

PR #379, opened 2025-07-02
Remove unnecessary throw

Bitslayn:
Removes an unnecessary throw in WorldAPI's getBlocks function. Minecraft already sorts positions.


UnlikePaladin:
This exists so that the error message is readable. Also vanilla doesn't do that across all versions

Bitslayn:
> This exists so that the error message is readable. Also vanilla doesn't do that across all versions

Understood

PR #384, opened 2025-07-20
Added reading avatar.jsonc

Bitslayn:


common/src/main/java/org/figuramc/figura/avatar/local/LocalAvatarFetcher.java
```diff
@@ -18,7 +18,7 @@
 import java.util.stream.Stream;
 
 /**
- * Navigates through the file system, finding all folders containing avatar.json
```
PoolloverNathan, line 21: > or

PoolloverNathan (commented):
seems good aside from grammar in one comment

UnlikePaladin:
Was this actually tested, json ≠ jsonc, and GSON which Minecraft uses does NOT support jsonc

Bitslayn:
avatar.json supports comments as the comments just get ignored. Any other jsonc feature has not been tested or implemented if they exist. (Which there aren't any other features of jsonc)

Bitslayn:
This PR only allows reading an avatar.jsonc file and nothing else

Bitslayn:
> Was this actually tested, json ≠ jsonc, and GSON which Minecraft uses does NOT support jsonc

I did test it yes. It's easy to test with json by just ignoring the warning messages by your IDE/Text editor about comments not being supported in json

Bitslayn:
image
image
Pay attention to the "Contributors" header in my authors

Bitslayn:
image
Also with /* */

Bitslayn:
But I could imagine this might not work on different Minecraft versions since it's implicit support

UnlikePaladin (changes requested):
If you want this to be accepted, rather than assuming MC will implicitly parse stuff with comments, strip the comments with regex or some other way.

Bitslayn:
o7 Got it, I'll see what I can do

Bitslayn:
> If you want this to be accepted, rather than assuming MC will implicitly parse stuff with comments, strip the comments with regex or some other way.

A regex pattern will not work because someone could use // inside quotes and that'd be captured. Could I not just set the GsonBuilder to lenient mode?

Bitslayn:
Actually I got an idea

Bitslayn:
Lmk if I need to rebase

Bitslayn:
Here's my tests, logs right after all the regex
image

UnlikePaladin (approved):

PR #393, opened 2025-09-01
Add Neoforge artifacts to build

Riftlight:
If statement needed because we only have neoforge on 1.20.2+

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,13 @@ jobs:
           name: Forge-Artifacts
           path: forge/build/libs/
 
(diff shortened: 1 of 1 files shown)


penguinencounter:
i'm not seeing why the conditional is necessary, couldn't you just target 1.20.2 instead?

Riftlight:
yeah that's arguably a better idea, i just wanted to do it this way because Neoforge has always felt like the modloader most likely to be dropped or delayed for a branch and i didn't want a PR to be necessary in that case.

UnlikePaladin (approved):

PR #397, opened 2025-09-09
add neoforge support in 1.21.8

baguchi:
Some parts were rewritten by the neoforge main core, so I manually separated some mixins for neoforge and fabric.
If it has bug. let's me know!


penguinencounter:
it works (it has the same issues as the fabric build with the player heads but that's expected and is probably not a loader-specific issue :P)

baguchi:
also I fixed custom armor render fix in neoforge and fix crash about sodium

PR #398, opened 2025-09-10
boxed collections

PoolloverNathan:


build.gradle
```diff
@@ -68,7 +68,6 @@ allprojects {
 
 	tasks.withType(JavaCompile) {
 		options.encoding = "UTF-8"
-		options.release = Integer.valueOf(java_version)
```
UnlikePaladin: why'd you remove this, this is important when compiling with a newer jdk for older versions

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -0,0 +1,252 @@
...
+            return this.parts.contains(p);
+        } else {
+            throw new LuaError("Expected argument of either modelpart or collection for 'PartCollection:has'");
+        }
+    }
+
+    // filter seems useful
+    @LuaWhitelist
+    @LuaMethodDoc("models.collection.filter")
+    public PartCollection where(@LuaNotNil LuaFunction arg) {
```
UnlikePaladin: I think where might be confusing, renaming it to filter might be better

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -658,16 +664,12 @@ public boolean getVisible() {
             aliases = "visible",
             value = "model_part.set_visible"
     )
+    @Override
```
UnlikePaladin, line 670: there's no spacing between functions in a bunch of places such as here

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1568,15 +1526,15 @@ public FiguraModelPart copy(String name) {
...
+    @LuaMethodDoc("model_part.collection")
+    public PartCollection getCollection(String name) {
         return collections.get(name);
     }
 
     @LuaWhitelist
-    @LuaMethodDoc("model_part.get_collections")
-    public Map<String, List<FiguraModelPart>> getCollections() {
-        return collections;
+    @LuaMethodDoc("model_part.collections")
```
UnlikePaladin: is it getCollections or collections, add one or both and make the method name consistent with it please

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -89,12 +93,14 @@ public FiguraModelPart(Avatar owner, String name, @Nullable String uuid, PartCus
...
+            for (String n: collections) builder.put(n, new ImmutableSet.Builder());
+            walkCollections(builder, collections);
+            builder.forEach((BiConsumer<String, ImmutableSet.Builder>) (key, parts) -> this.collections.put(key, new PartCollection(() -> owner.luaRuntime.typeManager, parts.build())));
         }
     }
 
-    private void walkCollections(Map<String, List<FiguraModelPart>> root, String names[]) {
+    private void walkCollections(Map<String, ImmutableSet.Builder> root, String names[]) {
         if (collectionInfo != null)
         for (byte b: collectionInfo)
```
UnlikePaladin: add indentation here

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -89,12 +93,14 @@ public FiguraModelPart(Avatar owner, String name, @Nullable String uuid, PartCus
         this.children = children;
         this.collectionInfo = collectionInfo;
         if (collections != null) {
-            for (String n: collections) this.collections.put(n, new ArrayList<>());
-            walkCollections(this.collections, collections);
+            var builder = new HashMap();
+            for (String n: collections) builder.put(n, new ImmutableSet.Builder());
+            walkCollections(builder, collections);
+            builder.forEach((BiConsumer<String, ImmutableSet.Builder>) (key, parts) -> this.collections.put(key, new PartCollection(() -> owner.luaRuntime.typeManager, parts.build())));
```
UnlikePaladin: this is kinda funky because they all parts get their own instance of a Collection with the shared set of Parts, memory usage goes 📈, ideally they'd share their instance by having the root create a Mutable collection of sorts that then gets frozen or something 🥴

UnlikePaladin (changes requested):
There's a few minor formatting issues and details, as well as a memory usage concern I have with how collections are walked, details in the comments, rest of the functionality looks pretty good!

build.gradle
PoolloverNathan: oops, forgot I removed it

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: I think there's only memory usage for the bbmodels, if I'm understanding my own code right?

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: I quickly renamed it and forgot to update the docs, oops

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: You collect Parts and store them in an Immutable set as the model is walked through which is fine, but then after you create a Set per Part (as you create a collection instance per part), this will make memory usage skyrocket for Models with many groups or nested groups

Bitslayn: What is this? Making collections accessable from all parts and not just the "bbmodel" groups?

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
PoolloverNathan: yeah, fair, been doing too much FP recently

PoolloverNathan: lol, I guess I went to rename it earlier and changed the doc, not the method name

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: I don't see the per-part allocation here? `collections != null` is only true on parts corresponding to bbmodels

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: I don't see myself introducing any spacing changes? that's an annotation directly above that, not a function

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: > What is this? Making collections accessable from all parts and not just the "bbmodel" groups?

No, collections are specifically only accessible from the bbmodel groups. We walk the part tree to collect parts in a collection, since parts store their collection indices.

Bitslayn: Okay, I don't see how that'd be different from making animations accessible from bbmodel groups which Figura already does

common/src/main/java/org/figuramc/figura/lua/api/AvatarAPI.java
```diff
@@ -31,7 +31,7 @@
 )
 public class AvatarAPI {
 
-    private final Avatar avatar;
+    final Avatar avatar;
```
penguinencounter: Not particularily against this, but there doesn't seem to be a reason for changing this to package-private

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
penguinencounter, line 1: should probably mention the method names instead of the operators in the error messages (ex. `PartCollection union` instead of `PartCollection +`)

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -89,15 +93,20 @@ public FiguraModelPart(Avatar owner, String name, @Nullable String uuid, PartCus
         this.children = children;
         this.collectionInfo = collectionInfo;
         if (collections != null) {
-            for (String n: collections) this.collections.put(n, new ArrayList<>());
-            walkCollections(this.collections, collections);
+            this.collections = new HashMap();
+            var builder = new HashMap();
+            for (String n: collections) builder.put(n, new ImmutableSet.Builder());
+            walkCollections(builder, collections);
+            builder.forEach((BiConsumer<String, ImmutableSet.Builder>) (key, parts) -> this.collections.put(key, new PartCollection(() -> owner.luaRuntime.typeManager, parts.build())));
```
penguinencounter: there are so many warnings here because of the lack of type arguments

Image

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -89,15 +93,20 @@ public FiguraModelPart(Avatar owner, String name, @Nullable String uuid, PartCus
...
+        } else {
+            this.collections = null;
         }
     }
 
-    private void walkCollections(Map<String, List<FiguraModelPart>> root, String names[]) {
+    private void walkCollections(Map<String, ImmutableSet.Builder> root, String names[]) {
```
penguinencounter: `ImmutableSet.Builder` has a type parameter, and it should be `String[] names` instead of `String names[]` (convention, not requirement)

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1569,14 +1530,14 @@ public FiguraModelPart copy(String name) {
 
     @LuaWhitelist
     @LuaMethodDoc("model_part.get_collection")
-    public List<FiguraModelPart> getCollection(String name) {
+    public PartCollection getCollection(String name) {
         return collections.get(name);
```
penguinencounter: `collections` is `@Nullable`, so this can NPE

common/src/main/java/org/figuramc/figura/lua/api/MutablePart.java
```diff
@@ -0,0 +1,222 @@
+package org.figuramc.figura.lua.api;
+
+import org.figuramc.figura.lua.LuaWhitelist;
+import org.figuramc.figura.lua.docs.LuaMethodDoc;
+import org.figuramc.figura.lua.docs.LuaMethodOverload;
+import org.figuramc.figura.math.vector.FiguraVec2;
+import org.figuramc.figura.math.vector.FiguraVec3;
+import org.figuramc.figura.model.FiguraModelPart;
```
penguinencounter: unused (`FiguraModelPart`) :)

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -0,0 +1,254 @@
...
+
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+@LuaWhitelist
+@LuaTypeDoc(name = "PartCollection", value = "models.collection")
+public class PartCollection implements MutablePart<PartCollection> {
+    private Supplier<LuaTypeManager> manager;
+    private Set parts;
```
penguinencounter: missing type argument

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
penguinencounter, line 1: throughout: `instanceof ModelPart` should be `instanceof FiguraModelPart` because `ModelPart` is the Minecraft type and not the Figura one

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    @LuaWhitelist
+    @LuaMethodDoc("models.collection.filter")
+    public PartCollection filter(@LuaNotNil LuaFunction arg) {
+        ImmutableSet.Builder<FiguraModelPart> builder = ImmutableSet.builder();
+        Iterable<FiguraModelPart> iter = parts;
+        for (var part: iter)
+        if (arg.invoke(manager.get().javaToLua(part)).toboolean(1))
+        builder.add(part);
```
penguinencounter: indentation

penguinencounter (changes requested):

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
UnlikePaladin: ^^^super important, thanks for spotting that

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1569,14 +1530,14 @@ public FiguraModelPart copy(String name) {
...
     }
 
     @LuaWhitelist
     @LuaMethodDoc("model_part.get_collections")
-    public Map<String, List<FiguraModelPart>> getCollections() {
-        return collections;
+    public Map<String, PartCollection> getCollections() {
+        return this.collections;
```
penguinencounter: intellij wants me to tell you that this should be `@Nullable`

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: ^^be sure to return nil here

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: nvm then

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: > I don't see the per-part allocation here? `collections != null` is only true on parts corresponding to bbmodels

Even if multiple parts share the same collection names and (logically) the same group of parts, this approach does not share the actual Set or PartCollection objects between different FiguraModelPart instances. 

So, if someone has deeply nested parts or many parts with overlapping collections, they'll end up with many separate (but possibly identical) Set and PartCollection objects in memory—one per part per collection. Which increases memory usage in repetitive models.

Rather than creating them like that, what I suggested is creating the collections initially outside of the part, and then adding the parts directly to the collection and finally freezing said collections outside of the figura model part, which could have the side benefit making them available from collections:listCollections(modelName) as a global to get all collections in a Model. 

The reason this will happen is because children also get their collections so collections != null will always be true

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: ^^this is important too, doing <> takes like 2 seconds

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: ^ c style arrays moment

UnlikePaladin (changes requested):
Look at Penguin's feedback, and the allocation thing

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: oops—but does @nullability matter for Lua?

common/src/main/java/org/figuramc/figura/lua/api/AvatarAPI.java
PoolloverNathan: ah, artifact from a test change

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
PoolloverNathan: oh shit, thanks

PoolloverNathan: …a hole in my test suite has been spotted

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
penguinencounter: no, but `collections` is private so java consumers (i.e. add-ons) will use this method and would benefit from the annotation

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: should be fixed now?

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: oh, hm. I'll look into that when I'm home

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
penguinencounter: everything's clear except that `ImmutableSet.Builder<FiguraModelPart>` can be inferred (`<>`)

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: You didn't actually change anything relating to this 💀

PoolloverNathan: not home yet

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -89,15 +93,20 @@ public FiguraModelPart(Avatar owner, String name, @Nullable String uuid, PartCus
         this.children = children;
         this.collectionInfo = collectionInfo;
         if (collections != null) {
-            for (String n: collections) this.collections.put(n, new ArrayList<>());
-            walkCollections(this.collections, collections);
+            this.collections = new HashMap<>();
+            Map<String, ImmutableSet.Builder<FiguraModelPart>> builder = new HashMap<>();
+            for (String n: collections) builder.put(n, new ImmutableSet.Builder<FiguraModelPart>());
+            walkCollections(builder, collections);
+            builder.forEach((key, parts) -> this.collections.put(key, new PartCollection(() -> owner.luaRuntime.typeManager, parts.build())));
```
UnlikePaladin, line 100: Now know you're not home yet so no worries, but if it helps Parts sharing the Set Builder, does not mean they'll share the same Set instance when built (which they should be sharing considering they're the same Collection). You can observe this in the sources for the builder.
Screenshot 2025-09-10 at 2 13 21 p m
Parts not sharing the map is fine because logically they might have different collections so no issue there. :P 
Safe travels home

UnlikePaladin (changes requested):

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: > The reason this will happen is because children also get their collections so collections != null will always be true

Empirically, this is false. Logs only show collection allocation happening for bbmodel collections. Maybe you're thinking about `collectionInfo`?

```diff
             for (String n: collections) builder.put(n, new ImmutableSet.Builder<FiguraModelPart>());
             walkCollections(builder, collections);
+            FiguraMod.LOGGER.info("Allocating builder for {}", name);
             builder.forEach((key, parts) -> this.collections.put(key, new PartCollection(() -> owner.luaRuntime.typeManager, parts.build())));
         } else {
+            FiguraMod.LOGGER.info("NOT allocating builder for {}", name);
             this.collections = null;
         }
```

```
[16:25:12] [ForkJoinPool.commonPool-worker-3/INFO] (Figura) NOT allocating builder for cube
[16:25:12] [ForkJoinPool.commonPool-worker-3/INFO] (Figura) NOT allocating builder for cube
[16:25:12] [ForkJoinPool.commonPool-worker-3/INFO] (Figura) NOT allocating builder for bone
[16:25:12] [ForkJoinPool.commonPool-worker-3/INFO] (Figura) Allocating builder for collection_test
[16:25:12] [ForkJoinPool.commonPool-worker-3/INFO] (Figura) NOT allocating builder for models
```

I'll look into this if you have an example of a pathological model, but I don't see what you're stating.

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: resolving this for the other thread

PoolloverNathan:
Current test script [and avatar](https://ipfs.io/ipfs/QmeDSCLex72qhv9qq4xUAn8HZbrxsZh1MYv6ZqNS6goyfZ)
```lua
print(("\n"):rep(15000))
print("Root collections", models:getCollections())
assert(not models:getCollections())
local collections = models.collection_test:getCollections()
print("Model collections", collections)
local collection = models.collection_test:getCollection "collection"
print("• collection", collection)
assert(collections.collection == collection)
local empty = models.collection_test:getCollection "empty"
print("• empty", empty)
assert(collections.empty == empty)
assert(collection:union(empty) == collection)
assert(collection:subtract(empty) == collection)
assert(collection:intersect(empty) == empty)
assert(empty:union(collection) == collection)
assert(empty:subtract(collection) == empty)
assert(empty:intersect(collection) == empty)
assert(not models.collection_test:getCollection "nonexistent")
print()
local filtered = collection:filter(function(p) return p:getName() == "cube" end)
print("• filtered", filtered)
assert(collection:has(filtered))
assert(filtered:has(empty))
assert(collection:has(collection))
assert(filtered:has(filtered))
assert(empty:has(empty))

local bone = models.collection_test.bone
local cube = models.collection_test.cube
assert(collection:has(bone))
assert(collection:has(cube))
assert(not filtered:has(bone))
assert(filtered:has(cube))
assert(not empty:has(bone))
assert(not empty:has(cube))

print(filtered:getParts())
filtered:setColor(1, 0, 0)
assert(models.collection_test.cube:getColor() == vec(1, 0, 0))
```

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
UnlikePaladin: Well after looking more closely at the implementation in the file, I guess children aren't meant to be aware of collections or which they belong to according to the bbmodel spec, but then again it's weird then that they have access to getCollections at all and a confusing design choice for a user

UnlikePaladin: Tbh I don't like that, but I guess with how the tree is structured there's no alternative other than implementing a CollectionsAPI but that'd be unnecessary I guess it's fine then

UnlikePaladin: ugh, why is there no differentiation between a model and a model part/group, this design sucks

UnlikePaladin: if it works like that it's fine by me, and is good to go now, not your fault, F's fault for the bad design I suppose. Will merge it now

UnlikePaladin (approved):

PR #400, opened 2025-09-10
Made render tasks copyable

UnlikePaladin:
Render tasks can now be copied by users with renderTask:copy(), calling copy on a ModelPart also copies Tasks on it by reference to match child behavior for parts. But they are copied by value if deepCopy is used.

### DIFF
--- a/common/src/main/java/org/figuramc/figura/model/rendertasks/SpriteTask.java
+++ b/common/src/main/java/org/figuramc/figura/model/rendertasks/SpriteTask.java
@@ -459,6 +459,27 @@ public class SpriteTask extends RenderTask {
     }
 
     @Override
+    public RenderTask copy() {
+        SpriteTask copy = new SpriteTask(name + "_copy", owner, parent);
+        copy.texture = texture;
+        copy.textureW = textureW;
+        copy.textureH = textureH;
+        copy.width = width;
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/figuramc/figura/model/rendertasks/SpriteTask.java
```diff
@@ -458,6 +458,27 @@ public List<Vertex> getVertices() {
...
+        copy.regionW = regionW;
+        copy.regionH = regionH;
+        copy.u = u;
+        copy.v = v;
+        copy.r = r;
+        copy.g = g;
+        copy.b = b;
+        copy.a = a;
+        copy.renderType = renderType;
+        copy.vertices.addAll(vertices);
```
penguinencounter: copying a ModelPart deepcopies the vertices (instead of linking the vertex objects), see [`FiguraModelPart.java:114-123`](https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java#L114-L123), this would be inconsistent with that

UnlikePaladin: thanks for spotting it

PR #402, opened 2025-09-11
fix: color colors

PoolloverNathan:


penguinencounter (approved):
yknow what sure. but also did you know `Colors.style` exists

PR #403, opened 2025-09-11
Feat: PRE_RENDER Event

Bitslayn:
pre_render is an event which allows for applying camera transformations through the RendererAPI the frame of, rather the next frame.

This PR applies to versions below 1.20 and up to 1.20.4 inclusively. A patch must be applied for 1.20.5 and above which I will make a PR for.

Credits go to Steph @superpowers04 for writing the event, all I did was fix instruction counting


Bitslayn:
Lol I failed to keep the indentation

PR #406, opened 2025-09-11
qol: treat .jars as .zip avatars

PoolloverNathan:
allows loading an avatar from a .jar, useful for e.g. bundling an avatar with a mod implementing server-side features for it

literally just changes the `*.zip` check to check for `*.zip || *.jar`


UnlikePaladin:
Yeah no 💀

PR #407, opened 2025-09-13
forcing model rendering in first person

soup587:
can be pretty handy as it means users don't have to create their own first person rendering system or use the firstpersonmodel mod (which can have some unwanted behaviors), and also provides a way of always having a full-framerate render event source


soup587:
todo: maybe change naming to be more clear (but not too long), maybe an automatic option for hiding head parented parts?

common/src/main/java/org/figuramc/figura/lua/api/RendererAPI.java
```diff
@@ -47,6 +47,9 @@ public class RendererAPI {
     @LuaWhitelist
     @LuaFieldDoc("renderer.render_hud")
     public boolean renderHUD = true;
+    @LuaWhitelist
+    @LuaFieldDoc("renderer.render_first_person")
+    public boolean renderFirstPerson = false;
```
Riftlight: Shouldn't this type and line 171's type should be changed to Boolean because they're nullable?

penguinencounter: more like the one on line 184 should be switched to the primitive `boolean`, because it's being used [directly in a `||` expression](https://github.com/FiguraMC/Figura/pull/407/files#diff-5d7a1bba1e409c620fcad708c29cbea3a4c414bfd6dd584d3e9a4cfd94ca8f4bR177) and i'm not sure what `null` `renderFirstPerson` is supposed to mean anyway

soup587: i was mostly referencing from the other boolean fields in rendererapi (and i've just realised booleans default to false so i didn't even need to = false it oops)

soup587:
don't think i can figure out a good way of doing the automatic head hiding thing (i am not smart) but i'm sure that can be figured out later

Bitslayn:
I tested this and it works well

Bitslayn:
Only thing maybe could be a first person render context for modelpart render events

soup587:
a

PoolloverNathan (approved):
yolo

soup587:
a

PR #409, opened 2025-09-14
feat: TextTask line spacing

Bitslayn:
# Changes:
* Added setting the line spacing to TextTasks using `spacing`/`setSpacing` methods (Defaults to 2)
* Added passing text line spacing into `getTextHeight`, and `getTextDimensions` functions

# Why?

Emojis exist and there's no way to decrease the spacing between lines of emojis to make them flush without additional code. Figura already handles splitting text inside TextTasks so it'd make sense for Figura to also allow for setting the spacing.

image
image


Bitslayn:
Now's the time to ask why Figura hasn't done this sooner? All the functions were there, tasks were written to render in this way. Was this feature cut?

UnlikePaladin (approved):

PR #412, opened 2025-09-14
fix: PRE_RENDER instruction counting

Bitslayn:
This one ACTUALLY fixes the issue (hopefully)

# The issue
Render instruction counting could sometimes never reset, causing the avatar to overflow instructions and error

# The fix
Instruction counting for pre_render and render events no longer overlap. Render can reset its own instructions and pre_render now has its own instruction object. When render instructions are reset, the pre_render instructions are read and added to the current render instructions.

### DIFF
--- a/common/src/main/java/org/figuramc/figura/avatar/Avatar.java
+++ b/common/src/main/java/org/figuramc/figura/avatar/Avatar.java
@@ -293,7 +293,8 @@ public class Avatar {
         if (scriptError || luaRuntime == null || !loaded)
             return;
 
-        render.reset(permissions.get(Permissions.RENDER_INST));
+        // render.reset(permissions.get(Permissions.RENDER_INST));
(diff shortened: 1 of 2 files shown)


Bitslayn:
Oops lol wrong button

UnlikePaladin:
You should probably drop the merge commit

UnlikePaladin:
You can set the head to the other's hash and then force push

Bitslayn:
I didn't know you could do that :think:

PoolloverNathan:
> I didn't know you could do that :think:

yeah, it's neat, or even just `git push REMOTE +6561e99:pre-render`

Bitslayn:
Tried `git push Bitslayn/Figura +a0e8cd3:pre-render` but it said `fatal: 'Bitslayn/Figura' does not appear to be a git repository`

PR #414, opened 2025-09-15
Placeholders for TextTasks

Bitslayn:
* Fixed the ${badges} placeholder so that it'd show dynamic badges
* Added ${name} placeholder to TextTasks
* Fixed `getTextHeight` and `getTextDimensions` functions
* Rolled back lineSpacing being different between Java and Lua side to avoid confusion


Bitslayn:
I miscounted the line spacing for the new getTextHeight and getTextDimensions commit so let me fix that

Bitslayn:
1 is default line spacing for both Lua and Java side

PR #415, opened 2025-09-15
Only check CANCEL_DAMAGE if damage event actually tries to cancel it

superpowers04:
Should probably be tested to make sure it works, I can't check at the moment


Bitslayn:
It'd be more readable if you returned early and didn't nest

PoolloverNathan (changes requested):
I have no veto power, but please use early returns

superpowers04:
Should be fixed, I made those last commits on mobile so I have no idea if they came out correct

superpowers04:
Sigh indentation is wrong, If I remember tomorrow, I"ll fix it

UnlikePaladin (approved):

PR #416, opened 2025-09-15
fix: Remove check for non-v4 UUIDs

PoolloverNathan:
Due to FSB, having non-v4 players with avatars is commonplace. For example, Carpet bots act like offline players, and couldn't get avatars. This PR also allows FSB to work properly on offline servers in gsneral.

The only changes I've made are removing NetworkStuff.checkUUID() and its usages. It compiles, but I haven't tested it yet.


penguinencounter:
this is not the problem

PoolloverNathan:
irrelevant for intended use case; see DMs for more information

PR #417, opened 2025-09-15
Remove networking port restrictions

penguinencounter:
why do these exist, again? they make it so much more annoying to work with local services because binding to low ports is a privileged operation, and you can only use two of them (??)

feel free to tell me if this is a horrible take

<sub>we're really running with the tiny PRs lately aren't we?</sub>


PoolloverNathan (approved):
> why do these exist, again?

people got scared after 0.1.3

PR #418, opened 2025-09-15
Use FSB for fetching users more often

penguinencounter:
this is mostly for equipping carpet bots, but the previous code exposed some weird behaviour where connecting with a non-FSB client would prevent anyone else from seeing your FSB equipped avatar

now, if FSB is connected, it tries to use that all the time

### DIFF
--- a/common/src/main/java/org/figuramc/figura/backend2/NetworkStuff.java
+++ b/common/src/main/java/org/figuramc/figura/backend2/NetworkStuff.java
@@ -323,8 +323,7 @@ public class NetworkStuff {
     }
 
     public static void getUser(UserData user) {
-        boolean fetchUser = fsb().isPlayerConnected(user.id) || FiguraMod.isOffline(user.id);
-        if (fsb().connected() && fetchUser) {
+        if (fsb().connected()) {
             fsb().getUserAndApply(user);
             return;
         }


PoolloverNathan (commented):
seems like a reasonable patch; does it work? (maybe have chloe test it in prod)

UnlikePaladin:
This looks wrong, the check exists for a reason

penguinencounter:
is the point of it to be able to have both avatars from FSB & avatars from the main backend?

( cc: @lexize )

UnlikePaladin:
Users are allowed to disable FSB on purpose, the behavior where connecting with a non-fsb client would result in it not showing is a bug, but removing the rest of the check has potential unintentional side effects such as allowing illegal copies to use it.

penguinencounter:
wdym by "illegal copies"?
offline accounts can already use FSB, that's not what `isOffline` does

lexize:
This check is used to just, not fetch avatar from FSB if player is not connected to FSB. Addons can send stuff to spoof (i think that's the right word) the connection state to FSB when it comes to carpet bots. So i think this check should stay there.

penguinencounter:
ok then

lexize:
What isOffline check does is it checks if player is connected at all, so if they are not, client will automatically fetch an avatar from FSB. The reason is player heads.

PR #419, opened 2025-09-15
getModVersion/getModName

soup587:
returns the version of a mod and its defined name respectively, from its id

(...why was mod versioning not exposed to the api in the first place?)


soup587:
on the fence on whether they should return an empty string or null, empty string is probably more convenient but null is probably more clear?

soup587:
oops

penguinencounter (commented):
these functions should probably return `nil` (`null`) if the mod isn't loaded, not `""`

soup587:
okay!!!

UnlikePaladin (approved):
yeah null makes more sense

PR #422, opened 2025-09-16
Fix FiguraListDocs forcing Configs to load before it should

superpowers04:


superpowers04:
I should note, this does conflict with #421
Not sure which solution is better but :woman_shrugging:

PR #423, opened 2025-09-16
getName, setName and getParent for Part Collections

superpowers04:


common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -22,12 +22,42 @@
 public class PartCollection implements MutablePart<PartCollection> {
     private Supplier<LuaTypeManager> manager;
     public final Set<FiguraModelPart> parts;
+    public final FiguraModelPart parent;
+    public String name;
```
PoolloverNathan: collections should be immutable
```suggestion
    public final String name;
```

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -22,12 +22,42 @@
...
+    @LuaMethodDoc("models.collection.get_name")
+    public String getName() {
+        return name;
+    }
+    @LuaWhitelist
+    @LuaMethodDoc("models.collection.set_name")
+    public PartCollection setName(String name) {
+        this.name = name;
+        return this;
+    }
```
PoolloverNathan: this method mutates a collection and should not be definable

PoolloverNathan (changes requested):
I'm also currently working on this exact thing, and the 'parent' terminology is incorrect (I'm opting for 'owner')

superpowers04:
> I'm also currently working on this exact thing, and the 'parent' terminology is incorrect (I'm opting for 'owner')

To me, parent makes more sense. The parent of the collection is the model that has the collection

superpowers04:
> > I'm also currently working on this exact thing, and the 'parent' terminology is incorrect (I'm opting for 'owner')
> 
> To me, parent makes more sense. The parent of the collection is the model that has the collection

Also, parent/children terminology is used all over Figura for Model Parts so to me it makes sense

PR #424, opened 2025-09-16
feat: make collections track owner & name

PoolloverNathan:
Collections now track their owner and name, including across set ops, pursuant to the following axioms:

* Compound/derived collections have the name of the leftmost involved collection.
* Derived collections have the same owner as their precedent.
* Compound collections have the same owner as their precedents iff they all share the same owner.
* Compound collections from multiple owners (including nil) have a nil owner.

resolves #423


common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1614,6 +1616,17 @@ public FiguraModelPart newPart(@LuaNotNil String name, String parentType) {
         return newer;
     }
 
+    @LuaMethodDoc(
+        value = "model_part.collect",
+        overloads = @LuaMethodOverload(
+            argumentNames = {"name", "parts§c..."},
```
penguinencounter: don't do this - documentation can be exported (ex. to FIGS) and this will just look bad in those cases

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1634,6 +1647,23 @@ public Object __index(String key) {
             case "preRender" -> preRender;
             case "midRender" -> midRender;
             case "postRender" -> postRender;
+            case "collect" -> new VarArgFunction() {
```
penguinencounter: instantiate this once as a static field and rely on the `self` first argument instead of `this`

(or at least instantiate it once per ModelPart instance, not _every single time it's used_)

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1614,6 +1616,17 @@ public FiguraModelPart newPart(@LuaNotNil String name, String parentType) {
...
 
+    @LuaMethodDoc(
+        value = "model_part.collect",
+        overloads = @LuaMethodOverload(
+            argumentNames = {"name", "parts§c..."},
+            argumentTypes = {String.class, FiguraModelPart.class}
+        )
+    )
+    public PartCollection collect(String name, FiguraModelPart... parts) {
+        return new PartCollection(name, FiguraModelPart.this, () -> owner.luaRuntime.typeManager, ImmutableSet.copyOf(parts));
```
penguinencounter: the current `this` is already `FiguraModelPart` here, no need to explicitly declare it (presumably you moved this code from the inner class?)

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1634,6 +1647,23 @@ public Object __index(String key) {
...
             case "postRender" -> postRender;
+            case "collect" -> new VarArgFunction() {
+                @Override
+                public Varargs invoke(Varargs args) {
+                    LuaTypeManager manager = owner.luaRuntime.typeManager;
+                    String name = args.checkjstring(2);
+                    // I'm sorry
+                    var parts = new FiguraModelPart[args.narg() - 2];
+                    for (int i = 2; i < args.narg(); i++) {
+                        if (manager.luaToJava(args.arg(i+1)) instanceof FiguraModelPart part) {
```
penguinencounter: just use `checkuserdata` instead of routing it through the type manager

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
```diff
@@ -20,20 +22,45 @@
...
     }
+    public PartCollection(String name, @Nullable FiguraModelPart owner, @Nullable FiguraModelPart owner2, Supplier<LuaTypeManager> manager, Set<FiguraModelPart> parts) {
+        this(name, owner == owner2 ? owner : null, manager, parts);
+    }
+    public PartCollection(PartCollection old, Set<FiguraModelPart> parts) {
+        this(old.name, old.owner, old.manager, parts);
+    }
+    public PartCollection(PartCollection left, PartCollection right, Set<FiguraModelPart> parts) {
+        this(left.name, left.owner, right.owner, left.manager, parts);
+    }
```
penguinencounter, line 48: these should probably have javadocs because it's not obvious what they do from a callsite

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: yes, correct — previously the method body was `throw new AssertionError();` but that'd be useless to e.g. Java callers

common/src/main/java/org/figuramc/figura/lua/api/PartCollection.java
PoolloverNathan: fair yeah I'll write those in the morning

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: oh yeah figs exists — so just remove the §c?

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
PoolloverNathan: …I forgot that exists okay

common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java
```diff
@@ -1644,6 +1675,7 @@ public void __newindex(@LuaNotNil String key, LuaFunction value) {
             case "preRender" -> preRender = value;
             case "midRender" -> midRender = value;
             case "postRender" -> postRender = value;
+            case "collect" -> _collectFunc = value;
```
penguinencounter: why is this something you're allowed to do

PoolloverNathan: idk; I just added it for consistency

PoolloverNathan: should I remove it?

PoolloverNathan: yeah it doesn't really make sense as a valid operation

penguinencounter: `collect` is acting like a method, not a field, so you shouldn't be able to set it

penguinencounter (approved):

PR #426, opened 2025-09-18
Add support for emojis in hovertext

Captain8771:
prevents needing to manually put unicode characters with the internal font

### DIFF
--- a/common/src/main/java/org/figuramc/figura/font/Emojis.java
+++ b/common/src/main/java/org/figuramc/figura/font/Emojis.java
@@ -3,6 +3,7 @@ package org.figuramc.figura.font;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.HoverEvent;
 import net.minecraft.network.chat.MutableComponent;
 import net.minecraft.network.chat.Style;
 import net.minecraft.resources.ResourceLocation;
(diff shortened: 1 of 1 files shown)


Bitslayn:
Does this filter badges and blacklisted emojis?

Riftlight:
Should it? The badges that are blacklisted never appear in hovertext.
Also, maybe this is a good time to consider removing the blacklist

Captain8771:
> Should it? The badges that are blacklisted never appear in hovertext. Also, maybe this is a good time to consider removing the blacklist

i think removing the blacklist entirely is a bit much, as it'd be possible to impersonate a dev/supporter/whatnot with no way to differentiate it from the real deal, but i agree that some things definitely shouldn't be on it, like :x:

Riftlight:
There's definitely an argument against it, but badges and emojis are distinguishable both by badges' hover text and supporter and dev badges look quite different from the nearest emoji

Captain8771:
> There's definitely an argument against it, but badges and emojis are distinguishable both by badges' hover text and supporter and dev badges look quite different from the nearest emoji

hovertext can be imitated by using the raw unicode character and font, but also i thought you meant removing the blacklist entirely, not just for emojis lol

tivvo:
> i think removing the blacklist entirely is a bit much, as it'd be possible to impersonate a dev/supporter/whatnot with no way to differentiate it from the real deal, but i agree that some things definitely shouldn't be on it, like ❌

like, the fact i cant use the star emoji is _really_ annoying considering everything as i didn't have a supporter icon till recently and any way to identify my own symbol and especially with debugging and profilers the ❌ emoji not being available is Very annoying

**at least** only apply the blacklist to list and chat only, as with text tasks it becomes annoying to do certain things

> Does this filter badges and blacklisted emojis?

also i don't think it's a necessity for this because nothing uses hovertext with the blacklisted emojis ¯\_(ツ)_/¯ the main reason from the necessity around the removal of the blacklisted emojis is to prevent fake errors / fake badges, and i feel like it should stay that way

PR #431, opened 2025-09-27
Remove PRE_RENDER render context

Bitslayn:
The pre_render context does not change. It runs once per frame and world_render doesn't pass a render context so it makes sense to not have a render context for pre_render.

The current issue this solves is pre_render currently returns "OTHER" as its context (provided by Avatar.json `renderMode.name()`), when in reality it should have returned "RENDER", though having a constant context doesn't make sense either.


Bitslayn:
Not important

PR #433, opened 2025-09-28
fix: RenderTask light and overlay

Bitslayn:
Fixes render tasks taking the light level and overlay of another avatar if they don't set their own


Bitslayn:
With this, tasks now properly take the light level of just their parent group rather than any group

PR #404, opened 2025-09-11
Fix :getMousePos() for proper correction for High DPI / Retina Mode Displays

tivvo:
as said, modifies the :getMousePos() position by the contentScaleFactor

```java
@LuaWhitelist
	@LuaMethodDoc("client.get_mouse_pos")
	public static FiguraVec2 getMousePos() {
		MouseHandler mouse = Minecraft.getInstance().mouseHandler;
		FloatBuffer xScale = BufferUtils.createFloatBuffer(1);
		FloatBuffer yScale = BufferUtils.createFloatBuffer(1);
		GLFW.glfwGetWindowContentScale( (long) Minecraft.getInstance().getWindow().getWindow(), xScale, yScale);
		return FiguraVec2.of(mouse.xpos() * xScale.get(0), mouse.ypos() * yScale.get(0));
	}
```

_post patch version of my extura fork with this fix, also works w/ res change_
tiv 2025-09-11 @ java

PR #436, opened 2025-09-29
revert #404 in exchange for :getWindowSize() fix

tivvo:
OKAY SO with the help of @auriafoxgirl 
we discovered that the mousePos fix actually ends up breaking alot of popular ui libs _(like panels)_, and it's just more efficent to return the proper window size over the corrected mouse pos

_also i just noticed the extura text ignore that_

---
1. this ended up fixing a few things, such as QDAA's scaling being made possible as it relies on the same ui tactics as figura and minecraft itself

image
<center><i>(this is with my new version, original + my bug fix in #404)</i></center>

---
2. also we added a note to i think :getMousePos() in docs to explain proper ui scaling because we thought it'd be useful
tiv@30-09-2025- DiscordCanary #Yaffle

---
3. `:getWindowSize()` now has an optional parameter to return the original result, in case of anything legacy or anyone in specific relying on the original window size for their mouse detection and what not, it's just `:getWindowSize(true)` 
tiv@30-09-2025- DiscordCanary #Alligatorgar

### DIFF
--- a/common/src/main/java/org/figuramc/figura/lua/api/ClientAPI.java
+++ b/common/src/main/java/org/figuramc/figura/lua/api/ClientAPI.java
@@ -235,10 +235,20 @@ public class ClientAPI {
     }
 
     @LuaWhitelist
-    @LuaMethodDoc("client.get_window_size")
-    public static FiguraVec2 getWindowSize() {
+    @LuaMethodDoc(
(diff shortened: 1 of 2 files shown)


tivvo:
Sorry translators I'm #Evil like that

PR #437, opened 2025-09-30
Avatar variable security stuff

penguinencounter:
## Features
- **Storing userdata isn't as big of a problem now**; userdata will have their metatables linked to the _destination_ avatar's metatables, rather than the _source_ avatar's metatables.
  - this does not fix the fact that storing an instance of `Animation` immediately destroys any and all security
  - functions are kinda covered? they only get wrapped to track ownership and make `getCaller` work; you can easily get around this by passing functions in tables or other containers
  - we unwrap functions if possible
- `table.deepcopy(table, metatableOptions, copyUserdata)`
- `avatar:getCaller() -> UUID (string)` returns the UUID of the avatar that is calling your functions

## Bug fixes
- Fixed usage of `length` in `ReadOnlyTable` which could be intercepted with the `__len` metamethod

## Things to get in before this is merged
- [x] Documentation


penguinencounter:
ready for review :)

penguinencounter:
@PoolloverNathan i'd like a review on this if you have the time just to make sure we didn't do anything obviously wrong

Bitslayn:
👎 Too invasive and getCaller only works when functions are wrapped. Maybe add a __eq to your wrapped functions?

Bitslayn:
I'm highly against this being added to Figura however

ChloeSpacedOut:
The issues fixed here are pretty minor so there's not much benefit, but also a lot of downsides with breaking inter-avatar communication. I think it's best to wait until postwrite (rererewrite) before introducing breaking changes like this

Captain8771:
this PR is the only way to know who called your functions unless you do some prompter/acceptor pattern like Bitslayn's Lift or don't expose functions at all like Vicky's PNP, both of which feel a bit overcomplicated. at the very least, i (and a few others) would like to see `avatar:getCaller()` be added, if nothing else.

(PS: `avatar:getCaller()` _does_ work when protection is disabled, idk what Bitslayn was doing in her testing)

Bitslayn:
> this PR is the only way to know who called your functions unless you do some prompter/acceptor pattern like Bitslayn's Lift or don't expose functions at all like Vicky's PNP, both of which feel a bit overcomplicated. at the very least, i (and a few others) would like to see `avatar:getCaller()` be added, if nothing else.
> 
> (PS: `avatar:getCaller()` _does_ work when protection is disabled, idk what Bitslayn was doing in her testing)

Yes, I deduced this after noting NKCAT's whitelist still works with both avatars disabling provide and consume protections.

envicious:
> this PR is the only way to know who called your functions unless you do some prompter/acceptor pattern like Bitslayn's Lift or don't expose functions at all like Vicky's PNP, both of which feel a bit overcomplicated. at the very least, i (and a few others) would like to see `avatar:getCaller()` be added, if nothing else.
> 
> (PS: `avatar:getCaller()` _does_ work when protection is disabled, idk what Bitslayn was doing in her testing)

I think avatar:getCaller() is a great addition for avatar interop, it'll make exposing functions so much safer, would benefit pre-existing protocols like patpat, and would make future interop easier to create.

penguinencounter:
alright let's give this another try

Bitslayn:
Ooh

PR #440, opened 2025-10-08
feat: Item lore and block name functions

Bitslayn:
This will need a 1.21 commit to add all functionality of BlockState:getName()


Bitslayn:
I'm just gonna drop this PR, not close it since it's genuinely a good feature to have in Figura.

PR #441, opened 2025-10-10
feat: Skull render improvements

Bitslayn:
Changes:
* Renames ITEM_ENTITY added in another 0.1.6 PR to GROUND to align with existing render contexts
* Adds FIXED render type for skulls rendered in the item frame
* Adds GUI render type for skulls rendered inside containers
* Made sure all skulls which render on an entity pass that entity through the skull render event

Fixes:
* Fixes GROUND context not applying to stacks of dropped items


Bitslayn:
Before and after of the fix
image
image

Bitslayn:
I'm just turning this commit into a PR that adds a bunch of skull render contexts because they use the same mixin

Bitslayn:
Btw, deleting the old mixin for the ITEM_ENTITY mode causes a validation error, so I kept it in.

tivvo:
SOMEONE PLEASE MERGE THIS SHIT NOW!!!!!

Bitslayn:
Lol

PR #442, opened 2025-10-13
feat: Watch symlinked scripts, reload the avatar

Bitslayn:
This commit allows for scripts that exist in symlinked folders to be watched and reload the avatar. This does not change what scripts can be loaded by the avatar.


Bitslayn:
Avatars load symlinked scripts, which are not watched with this PR. I'm going to write that functionality before reopening the PR

Bitslayn:
Nvm I can't get that to work lol

PR #446, opened 2025-10-19
ParseVec3 and ParseVec2 will not create a new vector by default

superpowers04:


penguinencounter (changes requested):
This may lead to unexpected mutations of the input vector in places that don't entirely make sense, as well as linkages that weren't present before.

These need to be found and switched to explicitly copy for this to be useful.

unexpected mutation:
- https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/model/rendertasks/SpriteTask.java#L374-L375
- https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java#L1009-L1010

vector is stored as reference:
- ~~https://github.com/FiguraMC/Figura/blob/main/common/src/main/java/org/figuramc/figura/model/rendertasks/RenderTask.java#L192-L193~~
- https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/lua/api/RendererAPI.java#L508
- ~~a bunch of stuff in `FiguraModelPart`~~

<sub>Note: neither of these lists is comprehensive.</sub>

superpowers04:
> This may lead to unexpected mutations of the input vector in places that don't entirely make sense, as well as linkages that weren't present before.
> 
> These need to be found and switched to explicitly copy for this to be useful.
> 
> unexpected mutation:
> 
> * https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/model/rendertasks/SpriteTask.java#L374-L375
> * https://github.com/FiguraMC/Figura/blob/1.20/common/src/main/java/org/figuramc/figura/model/FiguraModelPart.java#L1009-L1010
> 
> vector is stored as reference:
> 
> * https://github.com/FiguraMC/Figura/blob/main/common/src/main/java/org/figuramc/figura/model/rendertasks/RenderTask.java#L192-L193
> * https://github.com/FiguraMC/Figura/blob/main/common/src/main/java/org/figuramc/figura/lua/api/RendererAPI.java#L508
> * a bunch of stuff in `FiguraModelPart`
> 
> Note: neither of these lists is comprehensive.

Oh oops, my bad. In those cases you can just add a `true` to the arguments of the function. When I checked, I didn't see those usages

penguinencounter (approved):

penguinencounter (approved):

PR #449, opened 2025-10-23
feat: Allow for searching disconnected avatars

### LINKED ISSUE #429: You cannot change player permissions when they are not online
### Bug Description

I understand this is an overlooked thing rather than an actual bug, but it gets a little annoying. Essentially, when you try to change a player's permissions in the Permissions tab, you can only change permissions of players currently in your server. It would be great to have a toggle among the two other toggles to switch between showing all players, and showing only online players. Maybe even a third option to show only players that have ever joined the server you are currently in, if that's possible. This option could also only show up when searching.

### Reproduction Steps

Go to the Permissions tab, and search up a player that is not currently on the server. The player will not appear in the list.

### Log File

Not needed; nor would it provide any info

Bitslayn:
~~Fixes https://github.com/FiguraMC/Figura/issues/429~~

### DIFF
--- a/common/src/main/java/org/figuramc/figura/gui/widgets/lists/PlayerList.java
+++ b/common/src/main/java/org/figuramc/figura/gui/widgets/lists/PlayerList.java
@@ -185,11 +185,15 @@ public class PlayerList extends AbstractList {
             element.disconnected = false;
         }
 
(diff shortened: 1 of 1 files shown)


Bitslayn:
image

PR #450, opened 2025-10-24
fix: vectors.toCameraSpace

Bitslayn:
Fixed values returned by vectors.toCameraSpace() to match <1.21


Bitslayn:
All I did was flip values, but when I get the time, I'll need to see if there's any issues with scale

Bitslayn:
Didn't see any issues with camera bobbing which makes sense

UnlikePaladin (approved):

PR #456, opened 2025-11-26
1.20.6 → 1.21

penguinencounter:
ok porting this version forward doesn't seem that bad

- merge 1.20.6 into 1.21
- merge fsb_1.21 into 1.21
- fix compile errors
- fix runtime crashes
- tested:
  - Load some avatars
  - Upload avatar to FSB
  - Download avatar from FSB


starbornesoul:
thank you! :>

soup587:
i'll need to do more testing and actually grab the error later but i tried this the other day and it was erroring on startup trying to find a nonexistent mixin

UnlikePaladin:
the reason I haven't done this is because 1.20.6 has issues that need fixing such as incorrect lighting on the action wheel

PR #467, opened 2025-12-30
Additional Texture Changes

penguinencounter:
- [x] Allow omitting xywh box selection parameters on most methods, defaulting to "entire image"
- [x] Allow math operations between textures of different sizes, if the overflow settings on each permits
- [x] Allow resizing operations to target the same texture name as the source texture
- [x] Errors mid-operation (most commonly this is an out-of-texture-bounds error) will now rollback the entire operation instead of leaving the operation partially applied to the image
- [ ] Make better error messages


penguinencounter:
we expanded the test battery a bit for this PR

[TexturesTestReport.txt](https://github.com/user-attachments/files/24379438/TexturesTestReport.txt)

penguinencounter:
[testTextures2.lua.txt](https://github.com/user-attachments/files/24379452/testTextures2.lua.txt)

PR #469, opened 2026-01-07
Miscellaneous 1.20.4 NeoForge fixes

penguinencounter:
- the game can actually launch now
- fixed crash when opening chat

### DIFF
--- a/neoforge/src/main/resources/figura.mixins.json
+++ b/neoforge/src/main/resources/figura.mixins.json
@@ -5,15 +5,11 @@
   "minVersion": "0.8",
   "client": [
     "CommandSuggestionsMixinNeoForge",
-    "ClientPacketListenerMixin",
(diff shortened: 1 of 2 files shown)


neoforge/src/main/resources/figura.mixins.json
UnlikePaladin, line 1: ?? why the changes to the mixin json again?

penguinencounter: because those mixins don't exist...?

it was causing the game to crash on launch

UnlikePaladin: the hell, how did that get past the merge process

UnlikePaladin: okay then, thanks

PR #470, opened 2026-01-08
Rework #313

penguinencounter:
This reverts commit 231297a7f6c11b85e46784fdb0e6406b2e2a0e52 (aka PR #313).

Upload automation is not a good feature for base Figura because it strains the backend and could be abused to effectively stream 100KiB/s (not intended), per [Discord](https://discord.com/channels/1129805506354085959/1129815148190761081/1458610086355140638).


PoolloverNathan (changes requested):
I don't see why this can be abused any more than manually spamming the upload button could be, and it allows a mouse-free method of performing a relatively common action. Is there any evidence that this is being abused? Why are complaints about this only being raised a year later?

> could be abused to effectively stream 100KiB/s (not intended)

This could be done by automating `/figura load` and manually pressing the upload button. Should we remove all Figura commands too?

PR #473, opened 2026-01-08
/figura upload (option 3: just delete it)

penguinencounter:
i don't like the position i'm in here in the middle of this drama so i'm making three PRs with three options. someone else (probably paladin) can deal with figuring out which one is right. i'm done with this.

#471 #472 #473


penguinencounter:
we're doing this one apparently

PR #478, opened 2026-02-04
fix: Text height calculation

Bitslayn:
If the number of lines are 0, this could result in a -1


penguinencounter:
rebaaaseeee /lh

Bitslayn:
I did lol

Bitslayn:
Refined Github has a button that "updates" the branch by pulling from upstream rather than rebasing

PR #485, opened 2026-02-26
Custom permissive `Path` impl for relative `require` etc

penguinencounter:
with the advent of "interesting" chunk names like `anim_name keyframe (part 'bone', time 1.2s, ROTATION ?)` from the keyframe error improvement changes, `require` suddenly stopped working in keyframes, but only on Windows because [Windows is lame and doesn't allow the `?` character in paths](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions).

this PR fixes that by making a custom implementation of `Path` that applies Unix rules (`/` is a separator. that's it).

test if your build is affected:
```
load([[ require("it_works") ]], "oops?")()
```

**incorrect output:**
```
vm error: java.nio.file.InvalidPathException: Illegal char <?> at position 5: /oops?
```

**correct output:**
```
Tried to require nonexistent script "it_works"!
```

### DIFF
--- a/common/src/main/java/org/figuramc/figura/utils/PathUtils.java
+++ b/common/src/main/java/org/figuramc/figura/utils/PathUtils.java
@@ -7,16 +7,16 @@ import org.luaj.vm2.LuaValue;
 
 public class PathUtils {
 
-    public static Path getPath (String path) {
-        if (path.isEmpty()) return Path.of("/");
+    public static Path getPath(String path) {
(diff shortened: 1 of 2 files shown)


penguinencounter:
and yes this [has been seen in the wild](https://discord.com/channels/1129805506354085959/1476010659660566709/1476010659660566709)

PR #487, opened 2026-03-01
Fix crash with EntityCulling & the players arm not showing up in first person

april-ivy:
Fixes NPE crashes caused by figura$getEntityId returning null when EntityCulling bypasses extractRenderState and fixes arms not rendering in first person due to the prerender callback incorrectly cancelling vanilla arm rendering


UnlikePaladin (approved):
ty, greatly appreciated

PR #489, opened 2026-03-03
Fix crash when checking offline status in single player

### LINKED ISSUE #488: Crash joining Single Player world: FiguraMod.isOffline
### Bug Description

Minecraft 1.20.1, Fabric 0.17.2, Figura 0.1.5+1.20.1-a5f5d5d2
Full ModPack is Hexxy4 from the hex casting discord. I can give full mod list + versions if you really want me to here.

Client log is here, with screenshot of the crashy area, from the little I understand. https://mclo.gs/4oXvTfZ

Image

Confirmed the crash does not occur if Figura is disabled.
This only happens on one of my two single player worlds, and I tried with two different accounts. This did not happen when I connected to a multiplayer server. Any guidance is appreciated.

### Reproduction Steps

Load up Minecraft, join single player world, load up to 100%, see "Joining world...", _get crashed_.

### Log File

[4oXvTfZ.txt](https://github.com/user-attachments/files/25705563/4oXvTfZ.txt)

utafrali:
Fixes #488

The isOffline check was crashing in single player mode because the connection can be null. Added a simple null check before we try to access the online player list.

### DIFF
--- a/common/src/main/java/org/figuramc/figura/FiguraMod.java
+++ b/common/src/main/java/org/figuramc/figura/FiguraMod.java
@@ -1,6 +1,7 @@
 package org.figuramc.figura;
 
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.multiplayer.ClientPacketListener;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.Style;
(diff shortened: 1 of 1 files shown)


utafrali:
gonna close this for now, doesn't seem like it's needed anymore

PR #523, opened 2026-08-18
Add limit to string.rep function

xandercreates:
Prevent Memory Related Crashes from Stupidly long Repeating Strings.

### DIFF
--- a/common/src/main/resources/assets/figura/scripts/sandbox.lua
+++ b/common/src/main/resources/assets/figura/scripts/sandbox.lua
@@ -11,7 +11,7 @@ collectgarbage = nil
 local _original_rep = string.rep
 string.rep = function(s, n, sep)
     if n > 10000000 then
-        error("string.rep: too many repetitions (" .. n .. " > 100000) If you're seeing this, Congrats. You can't crash with strings reaching GiBs in length", 3)
+        error("string.rep: too many repetitions (" .. n .. " > 10000000), 3)
     end
     return _original_rep(s, n, sep)
 end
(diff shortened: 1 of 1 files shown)


common/src/main/resources/assets/figura/scripts/sandbox.lua
```diff
@@ -8,5 +8,13 @@ dofile = nil
 loadfile = nil
 collectgarbage = nil
 
+local _original_rep = string.rep
+string.rep = function(s, n, sep)
+    if n > 10000000 then
+        error("string.rep: too many repetitions (" .. n .. " > 10000000), 3)
```
penguinencounter: syntax error: `unfinished string near '" > 10000000), 3)'`

also this incurs non-zero overhead, i'll ask kathy to test this

penguinencounter (changes requested):

common/src/main/resources/assets/figura/scripts/sandbox.lua
xandercreates: oh whoops my bad

penguinencounter (commented):
try going a bit slower before committing

this is also invalid syntax _(you probably meant to put the close paren inside the string?)_

xandercreates:
:WHAT:

xandercreates:
Brain is not braining evidently, sorry

penguinencounter:
other thing i'll note is that this is only really good for preventing accidental calls to `string.rep` with very large repeat counts
if someone is intentionally trying to use `string.rep` to consume memory, they can just call it twice for some good ol' exponential growth

(i guess it would flag people for a little bit, until they change their scripts?)

xandercreates:
guess ill give the java side of things a try

xandercreates:
closing, easy to bypass. better to do it in java, personally tried adding it to java, failed miserably. Opening Issue instead

